PR #2, opened 2019-01-31
deps: add ngtcp2

danbev:
This commit add ngtcp2 to the node bulid.

I was not sure if ngtcp2 has been decided upon or not but if so this might be a starting point for adding it as a dependency. 

##### Checklist


danbev:
> Yep, this preempts the pr I was going to open next week! Which branch was this pulled from?

Sorry about that. I just though I'd save someone else going through the same thing. I just added ngtcp2 to https://github.com/danbev/quic/tree/quic_add_ngtcp2 (I did not pull from any branch).

jasnell:
It's no problem at all! I love being saved work!! :-D 

There is a branch specifically for draft -17 which is the current version of the spec. I'm not sure if master is the latest or that branch. We should verify before pulling in

danbev:
> It's no problem at all! I love being saved work!! :-D

Ah great :)

>There is a branch specifically for draft -17 which is the current version of the spec. 

I used master but I'll update this to use that branch instead as I believe master is on draft-15. I'll double check this.

danbev:
@jasnell I've updated the first commit to use the draft-17 branch. 

I added two more commit, one adding the ngtcp2 version to `process.versions`, and the other a very basic builtin just to verify that the symbols from ngtcp2 are available. 

This builds and all test pass locally for a normal build and when configured `--without-ssl`.

PR #8, opened 2019-02-03
quic: the start of a low level API 

jasnell:
**DO NOT MERGE**

This is the start of the native layer API for the QUIC implementation. Opening the initial PR now so it's easier to follow along and comment on. Will be updating this frequently over the next week or so.


3lcolmo:
Hi! 
I would like to know what is the state of the QUIC implementation in nodejs. I am currently trying to change my nodejs server over HTTPS to QUIC.
Thanks in advance!!

jasnell:
All @nodejs/quic ... this is ready to start reviewing. There's a very basic API and server implementation here with minimally working data flow in both directions. There's still a ton to do here and lots that just doesn't work or is untested, but this is already a lot to review. I'd like to get this PR landed then get the ngtcp2 update landed then iterate from there.

deps/ngtcp2/lib/includes/config.h
```diff
@@ -0,0 +1,39 @@
+
```
jasnell: note: This is necessary for building ngtcp2 on windows

deps/ngtcp2/lib/ngtcp2_log.c
```diff
@@ -25,7 +25,11 @@
 #include "ngtcp2_log.h"
 
 #include <stdio.h>
+
```
jasnell: Note: We'll want to upstream this change

jasnell: Actually, scratch that, it looks like it's been fixed in the updated lib

src/node_crypto.cc
```diff
@@ -481,6 +519,17 @@ void SecureContext::Init(const FunctionCallbackInfo<Value>& args) {
       min_version = TLS1_2_VERSION;
       max_version = TLS1_2_VERSION;
       method = TLS_client_method();
+    } else if (strcmp(*sslmethod, "TLSv1_3_method") == 0) {
```
jasnell: This will need to be refactored a bit now that TLS 1.3 support has officially landed in core. Will fix in a rebase after landing

lib/quic.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const {
+  createSocket
+} = require('internal/quic/core');
+
+createSocket.createSocket = createSocket;
```
danbev: It is not clear to me why we are setting this property. Should this be removed and the export be something like
```js
module.exports = {
  createSocket
}
```
And usage would become:
```js
const quic = require('quic');
const socket = quic.createSocket({ type: 'udp4', port: 1234 });
```

src/node_quic_session.cc
```diff
@@ -0,0 +1,2953 @@
...
+  QUICSESSION_CONFIG(V)
+#undef V
+}
+
+// Sets the QuicSessionConfig using an AliasedBuffer for efficiency.
+void QuicSessionConfig::Set(Environment* env) {
+  ResetToDefaults();
+  AliasedBuffer<double, Float64Array>& buffer =
+      env->quic_state()->quicsocketconfig_buffer;
+  uint64_t flags = buffer[IDX_QUIC_SOCKET_CONFIG_COUNT];
```
danbev: Should this be `buffer[IDX_QUIC_SOCKET_CONFIG_COUNT - 1];` perhaps?

test/parallel/test-quic-simple.js
```diff
@@ -0,0 +1,43 @@
...
+
+socket.on('session', common.mustCall((session) => {
+  console.log(session);
+
+  session.on('secure', () => {
+    // We can only open a stream after the handshake has completed
+    const uni = session.openUnidirectionalStream();
+    uni.write("I wonder if it worked.");
+    uni.end('test');
+  });
```
mcollina: From an API perspective, we should not wait for this event. In other terms, the `'secure'` event is not providing any additional behavior for a user.

test/parallel/test-quic-simple.js
```diff
@@ -0,0 +1,43 @@
...
+const socket = createSocket({ type: 'udp4', port: 1234 });
+
+socket.listen({ key, cert, ca });
+
+socket.on('session', common.mustCall((session) => {
+  console.log(session);
+
+  session.on('secure', () => {
+    // We can only open a stream after the handshake has completed
+    const uni = session.openUnidirectionalStream();
```
mcollina: I think we can have a friendlier name for this, like `createStream({ halfOpen: true })` or something similar.

deps/ngtcp2/lib/includes/config.h
mcollina: The space? lol

doc/api/quic.md
```diff
@@ -0,0 +1,539 @@
...
+* Type: {integer}
+
+The system file descriptor the `QuicSocket` is bound to.
+
+### quicsocket.listen([options][, callback])
+<!--YAML
+added: REPLACEME
+-->
+
+* `options` {Object}
```
mcollina: i think these options should be passed in when we create the socket.

deps/ngtcp2/lib/includes/config.h
jasnell: No, silly, the file

src/node_quic_session.cc
jasnell: No, there's an extra slot here to communicate which items are set and which are not, similar to how settings work in the http2 impl.

test/parallel/test-quic-simple.js
jasnell: Yeah, this is temporary until we get the signaling and state mechanism fully implemented inside the session and stream objects.

doc/api/quic.md
jasnell: Yeah, still working out how all of the options are going to work. Should be tackling that in the next iteration.

jasnell:
Going to get this landed today, then start another iteration. There are lots of TODO statements throughout the code if folks want to pick things up. Will likely do smaller PRs from this point forward

jasnell:
Landed!

PR #15, opened 2019-03-28
quic: implement QuicSession OnPathValidation

danbev:
This commit implements path validation (Section-8.2) of the
quic-transport specification.

When an endpoint does a connection migration it needs to verify that it
can communicate with the remote peer through the new local address (the
migrated address). The migrating endpoint would send a PATH_CHALLENGE to
the remote peer. After receiving a PATH_RESPONSE successfully validation
, which is handled by ngtcp2, the local address of the connection should
be updated and used for future communication.

This commit updates the local address in the cases of successful path
validation.

What should happen if path validation fails?
We have tried to do a connection migration but failed. Should this throw and
error to indicate this. If the migration happened because a switch from wifi to
3g/4g then the old local address is probably not useable anymore. This
is left as a todo at the moment to get feedback.

Refs:
https://tools.ietf.org/html/draft-ietf-quic-transport-19#section-8.2

##### Checklist


src/node_quic_session.cc
```diff
@@ -1014,12 +1014,20 @@ int QuicSession::OnRemoveConnectionID(
...
-    ngtcp2_conn* conn,
-    const ngtcp2_path* path,
-    ngtcp2_path_validation_result res,
-    void* user_data) {
-  // TODO(@jasnell): Implement this?
+int QuicSession::OnPathValidation(ngtcp2_conn* conn,
+                                  const ngtcp2_path* path,
+                                  ngtcp2_path_validation_result res,
+                                  void* user_data) {
+  auto session = static_cast<QuicSession*>(user_data);
```
jasnell:
```suggestion
  QuicSession* session = static_cast<QuicSession*>(user_data);
```

Just to keep it consistent

danbev: 👍

danbev:
Landed in ecd9f2e9b6630b9abf4a32d27894454c37ad833f.

gengjiawen:
Why is there is `*.c` added ? @danbev

danbev:
@gengjiawen That was a mistake, it did not notice it before now. Thanks for pointing that out. I've created https://github.com/nodejs/quic/pull/20 for removing it.

PR #17, opened 2019-04-01
quic: fix compiler and doc warnings

danbev:
This PR contains a number of commit to address compilation warnings and doc warnings. The main motivation for this is that at the moment it is a little difficult when upgrading the ngtcp2 dependency because we have to figure out which warnings are new (if any)

##### Checklist


jasnell:
Btw, the bump up to draft-19 seems to have broken things. I'll be working on getting that fixed up this week so that we're back to working. Word is the interop testing event in May is targeting draft-20. There apparently aren't many that are on 19 currently so we can't really do any interop testing just yet but we need to at least get it working again with the draft-19 ngtcp2 example server/client

src/node_quic_session.h
```diff
@@ -465,8 +465,8 @@ class QuicSession : public AsyncWrap {
   size_t ncread_;
   uint64_t tx_crypto_offset_;
 
-  BIO* enc_in_ = nullptr;   // StreamListener fills this for SSL_read().
-  BIO* enc_out_ = nullptr;  // SSL_write()/handshake fills this for EncOut().
+  // BIO* enc_in_ = nullptr;   // StreamListener fills this for SSL_read().
+  // BIO* enc_out_ = nullptr;  // SSL_write()/handshake fills this for EncOut().
```
jasnell: These can just be removed as they are not used at all

danbev:
> Btw, the bump up to draft-19 seems to have broken things. I'll be working on getting that fixed up this week so that we're back to working.

Sorry about that! 

> Word is the interop testing event in May is targeting draft-20

Do you have any links regarding this that I can read up on?

danbev:
Landed in 20b3a9abe2ca067fa85ba5239a661a927f758e91, 0ccc7616c597b25021a3b9089a1b5a480ec83816, 94f5fea89a44e8d38603d23994349a2c4ab359e2, 7e69d5c9c1b81d686248569543bdd931943656a0, 9f91ef0add487809b6d8f5dbd2ee21820e492a96, 93ceff1a80f288b46715993dd2372e9af2e0a111, and 888ace7b6af42ece4b54269a063988546ac54d11.

PR #22, opened 2019-04-06
Useless and redundant code

gengjiawen:
##### Checklist


src/node_quic_session.cc
```diff
@@ -3229,7 +3227,7 @@ int QuicClientSession::SendPendingData(bool retransmit) {
   Debug(this, "Sending pending data for client session");
   int err;
 
-  err = SendPacket();
+  SendPacket();
```
gengjiawen: I am not sure this change ? Should we check this err too ?

@jasnell @danbev

jasnell: Should generally always check but I'm planning to refactor this and much of the err return code handling in an upcoming iteration

gengjiawen: Okay, I will drop the commit then :)

jasnell:
Landed!

Issue #23, opened 2019-04-17
QuicBuffer


jasnell:
Ok, so the implementation is in a state where things are basically working but there is still a ton of work to do in order to get it complete. There's one major PR that I'll get landed later on today. The next big piece will be nailing down the implementation of `QuicBuffer` so the data flow from JS to C++ and back, with as little memcpy'ing as possible, works and is as efficient as possible. I have a few other things to work on over the next couple of days but I'll be working sporadically on the `QuicBuffer` implementation over the coming days and would love input on it.

The basic break down is this:

`QuicSocket` wraps the `uv_udp_t` handle and deals with sending packets. Unlike traditional UDP, QUIC includes error handling that requires packets to be re-transmitted until they are acknowledged by the peer. This means we cannot simply fire and forget a `uv_buf_t` and free it when the libuv callback is triggered. We have to keep the `ub_buf_t` around until ngtcp2 receives and processes the acknowledgement.

Data is sent at multiple places: 1) by QuicSocket itselr (e.g. version negotiation, close, errors, etc) ... 2) by QuicSession (and the QuicServerSession and QuicClientSession specializations) during TLS handshake and regular stream management, and 3) by QuicStream.

There are multiple data buffers at play. TLS handshake has it's own outgoing buffer, every QuicStream has it's own outgoing buffer. There are several places in the code where more transient buffers are used. It's fairly complicated but the bottom-line is that we cannot free buffers as soon as the libuv callback is triggered because we may need to retransmit those multiple times.

The current `QuicBuffer` implementation is modeled after the example code in ngtcp2, and it works reasonably well for what it does right now but it does not really do what we need it to do, particularly with regards to proper Stream API support callbacks when we can actually free data. It also depends heavily on memcpy at several places making it quite inefficient. It needs a better implementation.

It will likely be a few days before I can really dig in but I would love to open this bit up to others who may have some good ideas on how to handle the data flow here (ahem, @addaleax :-D .. since I love what you did with http2's internal data flow).

If someone wanted to start poking around with those bits, I'm happy to answer questions or help point in the right direction :-)

addaleax:
@jasnell I just reviewed #18, but I guess that was a bit late? I can try to go through the source here once and clean up some stuff in the next one or two days, if you want?

jasnell:
[...]

PR #24, opened 2019-04-17
quic: begin refactoring QuicBuffer

jasnell:
Start necessary improvements to QuicBuffer.

See: https://github.com/nodejs/quic/issues/23

/cc @addaleax


src/node_quic_buffer.h
```diff
@@ -0,0 +1,255 @@
...
+    Cancel();  // Cancel the remaining data
+    // Verify that all data has been purged, either through
+    // Consume or Cancel
+    CHECK_EQ(length_, 0);
+  }
+
+  // Transfers the state of this QuicBuffer to another
+  // instance, including ownership of the linked list
+  // of uv_buf_t instances. The destination must be
+  // an empty QuicBuffer.
```
addaleax: Does it make sense to implement this as a move constructor? The function source looks exactly like one :)

src/node_quic_buffer.h
```diff
@@ -0,0 +1,255 @@
...
+    CHECK_EQ(0, size_);
+    dest->head_ = head_;
+    dest->tail_ = tail_;
+    dest->size_ = size_;
+    dest->length_ = length_;
+    head_ = nullptr;
+    tail_ = nullptr;
+    size_ = nullptr;
+    length_ = nullptr;
+    dest->root_.reset(root_.release());
```
addaleax:
```suggestion
    dest->root_ = std::move(root_);
```

src/node_quic_buffer.h
```diff
@@ -0,0 +1,255 @@
...
+    return length_;
+  }
+
+  // The total number of buffers
+  inline size_t Size() {
+    return size_;
+  }
+
+  // Drain the remaining buffers into the given vector
+  inline size_t DrainInto(std::vector<uv_buf_t*>* list) {
```
addaleax: With a struct as small as `uv_buf_t` it’s not usually worth using pointer indirection, if optimization is the reason for doing this?

src/node_quic_buffer.h
```diff
@@ -0,0 +1,255 @@
...
+    tracker->TrackFieldWithSize("length", length_);
+  }
+  SET_MEMORY_INFO_NAME(QuicBuffer);
+  SET_SELF_SIZE(QuicBuffer);
+
+ private:
+  inline void Push(quic_buffer_chunk* chunk) {
+    size_++;
+    if (!tail_) {
+      root_.reset(std::move(chunk));
```
addaleax: `std::move()` doesn’t do anything for pointers … is this intentional here?

src/node_quic_buffer.h
jasnell: yep, was thinking the same.

src/node_quic_buffer.h
jasnell: ah right, artifact of a prior change really that I hadn't changed back yet.

src/node_quic_buffer.h
jasnell: Yeah, was still working out how I wanted this bit to work... the short version of the issue is this:

1. Whenever we call SendPacket on the `QuicSocket`, we need to hand it some N-number of uv_buf_t instances via the `QuicSocket`'s `SendWrap` class. Right now, passing those off to the `SendWrap` does a memcpy so that the `QuicBuffer` can be reset and can continue to be used while libuv does it's thing sending the data off to the UDP socket. Note that we only get ack callbacks from ngtcp2 for crypto and stream data packets, we don't get acks for other things (like version negotiation and other control packets). So, in some cases, we create the `SendWrap` with it's copy of the data then reset the `QuicBuffer` immediately, in other cases (specifically during the TLS handshake and within `QuicStream`), we create the `SendWrap` with it's copy of the data and only clear data out of the buffer as it is acknowledged. Currently this is done by maintaining a `std::deque` of `QuicBuffer` instances. This new implementation tries to take the approach of moving to a linked list of `quic_buffer_chunk` instances within a `QuicBuffer`. TLS Handshake would still have it's own `QuicBuffer` at the `QuicSession` level, and each `QuicStream` would still have it's own `QuicBuffer`.... hopefully that makes sense  :-D

So, I'm not 100% on any of this implementation at this point and want to find something that works as efficiently as possible with as little copying as possible, that will allow for retransmissions.

addaleax: Yeah, that sounds like you want to have a `std::vector<uv_buf_t>` and return `uv_buf_t` from `Head()` as well, if that’s possible…

jasnell:
@addaleax... I'm pulling in the comments you made on the other PR also...

src/node_quic_stream.h
```diff
@@ -101,12 +104,14 @@ class QuicStream : public AsyncWrap,
   int ReadStop() override;
 
   // Required for StreamBase
-  int DoShutdown(ShutdownWrap* req_wrap) override;
+  int DoShutdown(
```
gengjiawen: This can be in single line, as `SendPendingData` method.

src/node_quic_buffer.h
```diff
@@ -0,0 +1,261 @@
...
+    void* user_data_,
+    v8::Local<v8::Object> keep_alive_) :
+    buf(buf_),
+    done(done_),
+    user_data(user_data_) {
+    if (!keep_alive.IsEmpty())
+      keep_alive.Reset(keep_alive_->GetIsolate(), keep_alive_);
+  }
+
+  inline ~quic_buffer_chunk() {
```
gengjiawen: add `override` ?

src/node_quic_buffer.h
```diff
@@ -0,0 +1,261 @@
...
+
+class QuicBuffer : public MemoryRetainer {
+ public:
+  inline QuicBuffer() :
+    head_(nullptr),
+    tail_(nullptr),
+    size_(0),
+    length_(0) {}
+
+  inline QuicBuffer(QuicBuffer&& src) :
```
gengjiawen: Add `noexcept`.

src/node_quic_buffer.h
```diff
@@ -0,0 +1,261 @@
...
+    size_(src.size_),
+    length_(src.length_) {
+    root_ = std::move(src.root_);
+    src.head_ = nullptr;
+    src.tail_ = nullptr;
+    src.size_ = 0;
+    src.length_ = 0;
+  }
+
+  QuicBuffer& operator=(QuicBuffer&& src) {
```
gengjiawen: Add `noexcept`.

src/node_quic_buffer.h
```diff
@@ -0,0 +1,261 @@
...
+    this->~QuicBuffer();
+    return *new(this) QuicBuffer(std::move(src));
+  }
+
+  inline ~QuicBuffer() {
```
gengjiawen: add `override`.

src/node_quic_buffer.h
```diff
@@ -0,0 +1,261 @@
...
+      void* user_data,
+      v8::Local<v8::Object> keep_alive) {
+    Push(new quic_buffer_chunk(buf, done, user_data, keep_alive));
+  }
+
+  inline bool Pop(int status = 0) {
+    if (!root_)
+      return false;
+    std::unique_ptr<quic_buffer_chunk> root(std::move(root_));
+    root_.reset(root.get()->next.release());
```
gengjiawen: Use `root_ = std::move(root->next);`. 
Ref: https://clang.llvm.org/extra/clang-tidy/checks/misc-uniqueptr-reset-release.html

src/node_quic_session.cc
```diff
@@ -1396,10 +1400,10 @@ QuicSession::QuicSession(
...
 
   wrap->DefineOwnProperty(
-    env()->context(),
-    env()->state_string(),
-    state_.GetJSArray(),
-    v8::PropertyAttribute::ReadOnly);
+      env()->context(),
+      env()->state_string(),
+      state_.GetJSArray(),
+      v8::PropertyAttribute::ReadOnly);
```
gengjiawen: Elevate this namespace ?
Also should we check the returned value ?

src/node_quic_session.h
```diff
@@ -51,9 +51,11 @@ class QuicSessionConfig {
   QuicSessionConfig() {}
```
gengjiawen: `QuicSessionConfig() = default;`

src/node_quic_session.h
```diff
@@ -70,50 +72,53 @@ class QuicSession : public AsyncWrap {
...
-  QuicSession(QuicSocket* socket,
-              v8::Local<v8::Object> wrap,
-              crypto::SecureContext* ctx,
-              AsyncWrap::ProviderType provider);
+  QuicSession(
+      QuicSocket* socket,
+      v8::Local<v8::Object> wrap,
+      crypto::SecureContext* ctx,
+      AsyncWrap::ProviderType provider);
   virtual ~QuicSession();
```
gengjiawen: add `override`.

src/node_quic_session.h
```diff
@@ -221,266 +232,269 @@ class QuicSession : public AsyncWrap {
...
+      uint64_t max_streams,
+      void* user_data);
   static int OnExtendMaxStreamsBidi(
-    ngtcp2_conn* conn,
-    uint64_t max_streams,
-    void* user_data);
+      ngtcp2_conn* conn,
+      uint64_t max_streams,
+      void* user_data);
   static int OnExtendMaxRemoteStreamsUni(
```
gengjiawen: Seems not implemented. Also `OnExtendMaxRemoteStreamsBidi`

src/node_quic_session.h
```diff
@@ -698,16 +720,19 @@ class QuicClientSession : public QuicSession {
   SET_MEMORY_INFO_NAME(QuicClientSession)
   SET_SELF_SIZE(QuicClientSession)
 
-  int SetSession(SSL_SESSION* session);
+  int SetSession(
+      SSL_SESSION* session);
   v8::MaybeLocal<v8::Object> GetRemoteTransportParamsBuffer();
```
gengjiawen: Seems not implemented. and `GetSessionTicketBuffer` too.

src/node_quic_stream.h
```diff
@@ -41,40 +41,43 @@ class QuicStream : public AsyncWrap,
...
 
   QuicStream(
-    QuicSession* session,
-    v8::Local<v8::Object> target,
-    uint64_t stream_id);
+      QuicSession* session,
+      v8::Local<v8::Object> target,
+      uint64_t stream_id);
 
   virtual ~QuicStream();
```
gengjiawen: add `override`

src/node_quic_stream.h
jasnell: Planning to leave these as they are for now. Will fix up styles in a future iteration

src/node_quic_session.h
jasnell: Will handle these later on

jasnell: Actually... nevermind... can pull these out now

jasnell:
Landed

PR #25, opened 2019-04-23
quic: fix linting issues

danbev:
##### Checklist

- [x] `make -j4 lint` (UNIX) passes


lib/internal/quic/core.js
```diff
@@ -16,8 +16,8 @@ const { Duplex } = require('stream');
 const { isArrayBufferView } = require('internal/util/types');
 const {
   isIP,
-  isIPv4,
-  isIPv6,
+  isIPv4, // eslint-disable-line no-unused-vars
```
gengjiawen: Why not remove them since they are not used ?

PR #27, opened 2019-05-01
quic: fixup from last rebase

jasnell:
Fixes following the most recent update from nodejs/node master 

##### Checklist

### DIFF
--- a/test/parallel/test-quic-simple.js
+++ b/test/parallel/test-quic-simple.js
@@ -15,7 +15,7 @@ const createSocket = require('quic');
 
 const socket = createSocket({ type: 'udp4', port: 1234 });
 
-socket.listen({ key, cert, ca });
+socket.listen({ key, cert, /*ca*/ });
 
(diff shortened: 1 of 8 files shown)


jasnell:
the most recent rebase from master broke a few things. For some reason, the handshake now also fails against the ngtcp2 example client. Not entirely sure why yet, still investigating

test/parallel/test-quic-simple.js
```diff
@@ -15,7 +15,7 @@ const createSocket = require('quic');
 
 const socket = createSocket({ type: 'udp4', port: 1234 });
 
-socket.listen({ key, cert, ca });
+socket.listen({ key, cert, /*ca*/ });
```
jasnell: For some reason I've yet to determine, adding the `ca` in the to secure context causes a handshake failure in the ngtcp2 example client. It's not yet clear if that's a bug on our end or theirs.

jasnell:
Landed

PR #31, opened 2019-05-20
Iteration 6

jasnell:
Ongoing ...

##### Checklist


jasnell:
Ok... @nodejs/quic ... here's the current status:

The basic implementation is generally working with some caveats.

1. The JS QuicStream<->Internal QuicStream data flow is not quite 100% there. It works on Windows but the end-of-output and end-of-input bits aren't working consistently across platforms. (Have I mentioned before how much I dislike our Streams API and implementation). I'll be working on making this more consistent.
2. Error handling is largely non-existant right now.
3. I don't believe flow control is working properly and there's definitely no backpressure implemented yet
4. The whole mechanism around retransmission and idle timeouts needs to be completely refactored and made more efficient.
5. Right now, when a QuicSession is told to send data, it will attempt to send stream data for every known stream even if the stream does not have any data to send. This can be made more efficient to only send from streams we know have pending data to send.
6. We still need to implement mechanisms for cancelling streams in a reasonable way

What is working?

1. The TLS handshake is working very well.
2. We have keylog support that matches what just landed in master with the additional QUIC specific extensions that allow us to use wireshark to capture and decrypt traffic. :tada: :100:
3. ALPN negotiation, SNI, and transport params are all functional
4. QuicClientSessions will emit session ticket, id, and remote transport parameters which can be passed in when creating a new QuicClientSession to allow for connection resumption
5. We can open unidirectional and bidirectional streams from both the client and the server.

There's still a ton of work to do here and none of the http/3 bits have been implemented at all. The plan there is to have subclasses of both the C++ and JS core classes that implement the additional http3 semantics on top.

See the `test/parallel/test-quic-client-server.js` test for an example of the current API surface. The docs in quic.md have not yet been updated to match but I will hopefully get to that in this iteration.

At some point, we will want to implement a prioritization mechanism. Whereas http/2 has prioritization built into the protocol, QUIC does not, and relies on the application to provide prioritization. However, HTTP/3 *does* include prioritization that is modeled after the http/2 model. Right now, when sending data, we simply iterate through the streams in the order they were inserted into the internal map. Once we have the prioritization scheme worked out, this will need to become a priority queue of streams that actually have data to send. I'm not currently sure if the nghttp3 library will help with this, so we'll just have to see once we get to that.

The next major task I will have on this is to refactor the JS QuicStream data flow so that it works consistently across all platforms, then I will update the documentation, followed by working on the error handling and performance.

lib/internal/quic/core.js
```diff
@@ -1307,15 +1339,15 @@ class QuicStream extends Duplex {
...
     handle.onread = onStreamRead;
     handle[owner_symbol] = this;
     this[async_id_symbol] = handle.getAsyncId();
     this[kHandle] = handle;
     this.#id = id;
     this.#session = session;
-    this._readableState.readingMore = true;
   }
 
   get serverInitiated() {
```
appden: I recognize this is not changed (yet) for this PR, but this should strictly return a boolean (same for `unidirectional`).

jasnell:
Alrighty then... I have QuicStream working consistently across Linux and Windows now and flow control should be working. Next task will be to catch the documentation up to date and start on error handling. Here soon we'll need to start working on a comprehensive set of tests.

jasnell:
Documentation updated!

jasnell:
@nodejs/quic ... this is ready to land

jasnell:
Landed!

PR #41, opened 2019-07-03
Iteration 8

jasnell:
Ongoing...


jasnell:
Narrowing in on the issues I've been having with this.

1. Some of the ngtcp2 API calls cannot be called from within a ngtcp2 callback... so some of the timings have to be refactored when running user code that calls back into javascript. This impacts some of the state transitions.

2. Ideally, the QuicSession and QuicSocket will be torn down as quickly as possible without deferring to nextTick or setImmediate but there are some final packets that are not currently being serialized before the objects are destroyed, causing some trouble. 

I'm heading out on vacation in a few minutes so will get back on this once I'm back.

gengjiawen:
build fails on macOS 10.15, looks like data type not match.

```console
In file included from ../src/node_quic.cc:6:
In file included from ../src/node_quic_crypto.h:7:
../src/node_quic_session-inl.h:68:24: error: no matching function for call to 'max'
  max_crypto_buffer_ = std::max(max_crypto_buffer_, MINIMUM_MAX_CRYPTO_BUFFER);
```

jasnell:
Progress! QuicSocket and QuicSession lifecycle is more consistent and predictable. There's still a memory issue happening on Windows that I've yet to track down but things do not appear to be segfaulting at least. This PR currently includes commits to revert the upgrade to OpenSSL 1.1.1c due to a bug introduced in random number generation.

test/parallel/test-quic-client-server.js
```diff
@@ -30,8 +30,11 @@ const filedata = fs.readFileSync(__filename, { encoding: 'utf8' });
```
gengjiawen: on Line 255: is `req.authenicated` a typo ?

jasnell: Yep! sure is.

gengjiawen:
Build failed with
```console
../src/node_quic_socket.cc:725:25: error: no matching member function for call to 'DrainInto'
  size_t len = buffer_->DrainInto(&vec, &length_);
               ~~~~~~~~~^~~~~~~~~
../src/node_quic_buffer.h:285:17: note: candidate function not viable: no known conversion from 'size_t *' (aka 'unsigned long *') to 'uint64_t *' (aka 'unsigned long long *') for 2nd argument
  inline size_t DrainInto(
                ^
../src/node_quic_buffer.h:306:17: note: candidate function not viable: no known conversion from 'std::vector<uv_buf_t> *' (aka 'vector<uv_buf_t> *') to 'std::vector<ngtcp2_vec> *' for 1st argument
  inline size_t DrainInto(
                ^
In file included from ../src/node_quic_session.cc:9:
In file included from ../src/node_quic_crypto.h:7:
In file included from ../src/node_quic_session-inl.h:10:
```

jasnell:
Got this landed, will fix the build failure in the next iteration

jasnell:
Fixup PR! https://github.com/nodejs/quic/pull/43

PR #44, opened 2019-07-19
deps: update to ngtcp2 draft-22

jasnell:
Begin update to ngtcp2 draft-22.

There are significant API changes in the underlying library here that simplify things overall but currently break the handshake. I'm still working through those breaks at the moment so this is WIP until I get those resolved.

##### Checklist


jasnell:
Alrighty... progress... handshake is working with fewer steps required, which is awesome. There are a number of other API changes and clarifications that have been made in draft-22 that are very helpful, but are breaking things in the current implementation (due to problems on our side). I'll be fixing those up next.

jasnell:
Additional progress made... still some failure.

jasnell:
Just to give folks an idea of the full list of things still being worked on:

1. openStream() API ... specifically, ability to specify an initial chunk of data to include when opening the stream.
2. Currently, in most cases, setting stream fin requires an empty stream frame with fin bit set. See if we can optimize that further
3. Avoid stream commitment attacks. Verify that implicitly opened streams do not commit resources unless data is actually sent.
4. Verify that STREAM_RESET and STOP_SENDING are handled appropriately
5. Ensure that close and destroy are re-entrant tolerant
6. Ensure that all C++ objects can be destroyed immediately in the current callstack without requiring nextTick or setImmediate deferral.
7. Ensure that only CONNECTION_CLOSE frames are sent in the closing state, Ensure that absolutely no frames in the draining state.
8. Ensure that STOP_SENDING is not sent after receiving a STREAM_RESET from the peer
9. Revisit the flow control / backpressure strategy. We may want to allow multiple flow control strategies to allow performance to be tuned.
10. Make use of final size parameter when known (at the very least, there should be a getter on the QuicStream)
11. Implementing a pool of connection IDs to make generation more efficient
12. Determine whether we want to support zero-length connection IDs. This is optional by the specification
13. Determine whether we want to expose an API to force a QuicSession to use a new connection ID. Changing the CID is permitted by the specification, we just need to decide whether to expose an API for it. Use of such an API won't be common but there is an argument to be made about having it available for privacy/security reasons.
14. We need an API to trigger a key update
15. Support ability to mark the QuicSession busy and unable to accept new connections (SERVER_BUSY response)
16. Ensure that a server completely ignores invalid or unexpected frames
17. Determine how and if we ever want to support multiple QUIC protocol versions
18. Verify how sending of version negotiation packets is handled. e.g. are they sent for 0RTT packets? 
19. alpn is absolutely required so ensure that it's always there. Error if not provided
20. Improve handling of configuration and transport params
21. Ensure that crypto handshake buffer protections are working
22. Ensure that crypto handshake retransmission is occurring properly
23. Wait for NEW_TOKEN support to be implemented by ngtcp2
24. Only prefer address validation once per connection per address? Make this configurable
25. Ensure that path validation is occurring correctly and is tested
26. Connection migration test is currently broken, need to verify that it's working and ensure completed implementation
27. Complete support for server preferred address. Upon receiving and validating a server preferred address, client should migrate to the preferred address automatically.
28. Ensure that moving to a server preferred address and migration to a new connection can actually happen simultaneously
29. Implement stateless reset support
30. Determine whether we should allow disabling the idle timeout at all. There are many reasons why we shouldn't allow disabling it.
31. Implement exponential backoff on the sending of CONNECTION_CLOSE frames during closing period.
32. Ensure application close on QuicSessions
33. Implement handling of an out of packet numbers error. That is, the QuicSession can run out of new packet numbers and fail to serialize new packets, when this happens, the connection must be torn down. We should implement a special error for this so that this case can be detected.
34. Verify that retransmissions are happening correctly
35. Verify ngtcp2 support for ECN
36. Determine whether we ever want to try support for path maximum transmission unit discovery. Is there anything we need to do to handle packets too large for MTU?
37. Use static allocated packet buffers rather than malloced
38. Verify ngtcp2 support for ICMP packet too big messages
39. Determine if we want to support latency spin bit in any way? Verify that ngtcp2 even allows us to
40. Determine api for supporting extension transport parameters
41. Determine if there's any reason to expose an API for ping frames. Determine how we could ever support responses.
42. Determine if there's any way to support extension frames with ngtcp2
43. Ensure proper handling and use of QUIC session and crypto error codes
44. Implement various slow loris style DOS mitigations
45. Verify amplification attack mitigations
46. Verify stream commitment attack mitigations

jasnell:
Landed

jasnell:
Moved the checklist of todos to https://github.com/nodejs/quic/issues/45

Issue #37, opened 2019-06-18
ping


splitice:
An interface to send QUIC ping frames would be really handy for long lived connections. It's also common in the existing QUIC prototype libraries that exist for javascript.

jasnell:
We certainly could, but handling the ping responses are a bit problematic in that they come in as normal Acks that are not surfaced via the ngtcp2 API. This may need to wait until ngtcp2 provides the necessary API bits. That said, ngtcp2 does keep rolling RTT stats that we do have access to.

splitice:
Yes looking the provided API there could be some complexity here. However it's reasonably common.

For context we are experimenting with porting an existing protocol implemented on a basic pure-js QUIC implementation (without TLS) to this implementation. This was a feature provided by that implementation, however for our usage we could easily do it at the stream layer.

jasnell:
So, investigating further on this, there are ways we can *send* a ping but ngtcp2 does not provide any mechanism for knowing that a ping frame has been acknowledged. Before going much further on this, we should agree on what the actual use case need here is.. specifically is the ping intended to:

1. Detect whether the connection is still active
2. Keep the connection active to avoid timeouts
3. Measure the RTT latency for the connection

For 1, with QUIC, a connection is only active when there's data actually being transmitted. There may be periods of dormancy that are limited by the idle_timeout on configuration. We can easily report on the gaps of time between activity as we are actively tracking that already (e.g. "there's been no activity in the past 20 milliseconds"). We don't need PING support for this.

For 2, we can send PING frames in a fire-and-forget type of way with a hope that it'll keep the connection open but there's really no guarantee. If we don't need guarantees, then just having the API to send a PING is sufficient.

For 3, ngtcp2 reports rolling RTT statistics that are updated every time frames are processed. We don't have to send a PING to tap into these metrics. We can simply surface them on every QuicSession instance for applications to tap into any time they want.

Are there any other use cases that we may need to cover?

splitice:
Yes that describes the three reasons we send pings. Right now to get our port of that application working we ended up with a quic stream kept open doing 1 byte req & responses to imitate that behavior (not efficient, but workable).

jasnell:
That's not an unreasonable approach to be honest, and it's something that we could make more efficient by implementing at the native layer.

PR #47, opened 2019-07-31
quic: add beginnings of a ping API

jasnell:
Beginnings of a simple PING api.. see: https://github.com/nodejs/quic/issues/37

This adds a single `ping()` method to `QuicSession` that will effectively just trigger serialization and sending of any frames currently pending in the ngtcp2_conn queue (this will include probe packets to test for connection liveness).

Unlike `ping()` in http/2, there is no response to ping in QUIC that we can monitor.

The QUIC `ping()` method is used only to keep the connection alive while we're not transmitting any stream packets, etc. The idle and retransmit timeouts will be updated and (currently) any errors that may occur trying to send the frames will be ignored, but that may change in the future.

##### Checklist

### DIFF
--- a/doc/api/quic.md
+++ b/doc/api/quic.md
@@ -315,6 +315,19 @@ Returns a new `QuicStream`.
 An error will be thrown if the `QuicSession` has been destroyed or is in the
 process of a graceful shutdown.
 
+### quicsession.ping()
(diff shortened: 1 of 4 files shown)


jasnell:
Landed

PR #49, opened 2019-07-31
quic: iteration 13

jasnell:
This iteration is really just a number of cleanups plus some additional error handling.

##### Checklist

### DIFF
--- a/src/node_quic_session.cc
+++ b/src/node_quic_session.cc
@@ -325,9 +325,7 @@ void QuicSession::Destroy() {
   // that pointer, allowing the QuicSession to be
   // deconstructed once the stack unwinds and any
   // remaining shared_ptr instances fall out of scope.
-  std::shared_ptr<QuicSession> ptr = this->shared_from_this();
   RemoveFromSocket();
-  ptr.reset();
 }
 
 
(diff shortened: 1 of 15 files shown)


src/node_quic_session.cc
```diff
@@ -243,39 +247,58 @@ void QuicSession::AssociateCID(ngtcp2_cid* cid) {
...
+
+  HandleScope scope(env()->isolate());
+  Context::Scope context_scope(env()->context());
+
   Local<Value> argv[] = {
     Number::New(env()->isolate(), static_cast<double>(last_error.code)),
     Integer::New(env()->isolate(), last_error.family)
   };
+
+  // Grab a shared pointer to this to prevent the QuicSession
```
trivikr: Looks like this call is missed on line 994 https://github.com/nodejs/quic/blob/63b1b2f81e92b91ebdb777baad77a48fb715e640/src/node_quic_session.cc#L989-L995

jasnell: @trivikr ... please take a look again. The line numbers don't appear to be matching up so I'm not sure what you're referring to here.

trivikr: I searched for calls of `MakeCallback(` in `node_quic_session.cc` and noticed that there are 15 of them https://github.com/nodejs/quic/blob/master/src/node_quic_session.cc

And noticed that following line is added only on 14 of them in this PR
`std::shared_ptr<QuicSession> ptr(this->shared_from_this())`

Line number:
* nodejs:master is 994
* jasnell:iteration-13 is 1064
https://github.com/jasnell/quic/blob/3b8a9efce8640ce49365d74503528ed0d9b2c5fb/src/node_quic_session.cc#L1059-L1065

jasnell: Just did a quick scan through and verify that all of the MakeCallback calls in node_quic_session.cc include shared_ptr

jasnell:
This PR should be ready to go

jasnell:
Landed!

PR #53, opened 2019-08-21
quic: fix warning for ClearTLS parameter

danbev:
Currently the following compiler warning is generated:
```console
../src/node_quic_session.cc:1494:26: warning:
logical not is only applied to the left hand side of this comparison
[-Wlogical-not-parentheses]
  return ClearTLS(ssl(), !Type() == QUICSESSION_TYPE_SERVER);
                         ^       ~~
../src/node_quic_session.cc:1494:26: note:
add parentheses after the '!' to evaluate the comparison first
  return ClearTLS(ssl(), !Type() == QUICSESSION_TYPE_SERVER);
                         ^
                          (                                )
```
This commit changes the expression to use the not equal to operator.

##### Checklist


src/node_quic_session.cc
```diff
@@ -1491,7 +1491,7 @@ int QuicSession::TLSHandshake() {
 // read it and throw it away, unless there's an error.
 int QuicSession::TLSRead() {
   ClearTLSError();
-  return ClearTLS(ssl(), !Type() == QUICSESSION_TYPE_SERVER);
+  return ClearTLS(ssl(), Type() != QUICSESSION_TYPE_SERVER);
```
jasnell: Heh... sigh. That must have been a late in the evening edit originally 😆

jasnell:
Landed!

PR #84, opened 2019-08-22
quic: add check for nullptr in SeekHead

danbev:
Currently, when the next quic_buffer_chunk is null SeekHead will cause a
segment fault as it tries to accesss head_->buf.len, but head_ will be
null at this point.

This commit attempts to fix this by adding a check for this situation
and setting rlength to zero.

##### Checklist


jasnell:
landed

Issue #67, opened 2019-08-21
Finish SERVER_BUSY support


jasnell:
The basic mechanism for SERVER_BUSY is in place but needs to be finished pending completed support for sending CONNECTION_CLOSE in an initial packet.

jasnell:
Pending resolution of https://github.com/ngtcp2/ngtcp2/issues/145

trivikr:
Unblocked as it is allowed to send CONNECTION_CLOSE frame in Initial packet as long as ngtcp2_conn has initial key installed and it is not in closing or draining state. https://github.com/ngtcp2/ngtcp2/issues/145#issuecomment-523765371

jasnell:
The current implementation I have in the PR will go ahead and create a `QuicServerSession` but will immediately cause it to abort processing the initial packets once the initial keys are installed, then will send the connection close. It's not ideal in that it makes things a bit more expensive when we know we're not going to be doing any work. I'd rather avoid the calls across to the JavaScript side entirely, so we'll need to revisit the implementation on it but for now the basic mechanism works.

jasnell:
Closing as the current mechanism works even if not ideal

PR #85, opened 2019-08-22
quic: implement SERVER_BUSY

jasnell:
Initial server busy handling. Unfortunately there's going to be
a little bit of overhead when rejecting a connection because we
have to install the keys, but we need to attempt to minimize it.

Should explore how to minimize it further. Maybe we should have
a special QuicSession specialization that only handles rejections?

##### Checklist


trivikr:
Refs: #67

jasnell:
Landed

PR #102, opened 2019-08-29
quic: use a unique-ptr for Timer

jasnell:
Alternative to https://github.com/nodejs/quic/pull/100

/cc @danbev

##### Checklist

### DIFF
--- a/src/node_quic_session.h
+++ b/src/node_quic_session.h
@@ -860,8 +860,8 @@ class QuicSession : public AsyncWrap,
   size_t connection_close_attempts_ = 0;
   size_t connection_close_limit_ = 1;
 
-  Timer* idle_;
-  Timer* retransmit_;
+  TimerPointer idle_;
+  TimerPointer retransmit_;
 
   CryptoContext hs_crypto_ctx_{};
   CryptoContext crypto_ctx_{};
--- a/src/node_quic_util.h
+++ b/src/node_quic_util.h
@@ -490,7 +490,10 @@ class Timer {
     timer->env_->CloseHandle(
         reinterpret_cast<uv_handle_t*>(&timer->timer_),
         [&](uv_handle_t* timer) {
-          delete timer;
+          Timer* t = ContainerOf(
+              &Timer::timer_,
+              reinterpret_cast<uv_timer_t*>(timer));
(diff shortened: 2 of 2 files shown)


danbev:
I'm sure I'm missing something here, but I don't see the need to do any memory clean up apart from what is done by the unique_ptr, which will clean up the Timer instance when the QuicSession destructor is called. 

`uv_timer` is not a pointer in Timer class:
```console
(lldb) fr v timer->timer_
(uv_timer_t) timer->timer_ = {
  data = 0x0000000000000000
  loop = 0x00000001029f9170
  type = UV_TIMER
  close_cb = 0x0000000000000000
  handle_queue = ([0] = 0x00000001029f9180, [1] = 0x0000000104f420f0)
  u = {
    fd = 0
    reserved = ([0] = 0x0000000000000000, [1] = 0x0000000000000001, [2] = 0x0000000000000000, [3] = 0x0000000000000000)
  }
  next_closing = 0x0000000000000000
  flags = 8
  timer_cb = 0x0000000000000000
  heap_node = ([0] = 0x0000000000000000, [1] = 0x0000000000000000, [2] = 0x0000000000000000)
  timeout = 0
  repeat = 0
  start_id = 0
}
```
When the the callback registered in `Timer::Free` is called, it will try to delete something that has not been allocated leading to the following error:
```console
node(31513,0x104c3c5c0) malloc: *** error for object 0x10690bbd0: pointer being freed was not allocated
node(31513,0x104c3c5c0) malloc: *** set a breakpoint in malloc_error_break to debug
```

jasnell:
doh, you're right, freed the wrong thing ;-) ... The issue here is that the uv_timer_t handle needs to be closed when it is deconstructed, and that does not necessarily happen synchronously. Instead of freeing the uv_timer_t tho, I needed to free the Timer itself in the callback. Will update the PR in a minute

danbev:
> The issue here is that the uv_timer_t handle needs to be closed when it is deconstructed, and that does not necessarily happen synchronously

Ah I see, thanks for clarifying that!

danbev:
Landed in 362d112b389250c97c6e6f2c6d58305b2347d6f1.

PR #104, opened 2019-09-01
quic: updating crypto and handshake impl

jasnell:
This is preparation for updating ngtcp2 to the latest version, which completely reworks the crypto/tls APIs. I was going to bundle everything into a single PR but there are just too many changes and it's better to split things out like this...

##### Checklist

### DIFF
--- a/src/node_quic_crypto.cc
+++ b/src/node_quic_crypto.cc
@@ -1,0 +1,1809 @@
+#include "node_quic_crypto.h"
+#include "env-inl.h"
+#include "node_crypto.h"
+#include "node_quic_session-inl.h"
+#include "node_quic_util.h"
+#include "node_url.h"
+#include "string_bytes.h"
+#include "v8.h"
+
+#include <ngtcp2/ngtcp2.h>
+#include <openssl/bio.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include <openssl/kdf.h>
+#include <openssl/ssl.h>
+#include <openssl/x509v3.h>
+
+#include <iterator>
+#include <numeric>
+#include <unordered_map>
+#include <string>
+#include <sstream>
+
+namespace node {
+
+using v8::Array;
+using v8::Local;
+using v8::Object;
+using v8::String;
+using v8::Value;
+
+namespace quic {
+
+namespace {
+int BIO_Write(BIO* b, const char* buf, int len) {
+  return -1;
+}
+
+int BIO_Read(BIO* b, char* buf, int len) {
+  BIO_clear_retry_flags(b);
+  QuicSession* session = static_cast<QuicSession*>(BIO_get_data(b));
+  len = session->ReadPeerHandshake(reinterpret_cast<uint8_t*>(buf), len);
+  if (len == 0) {
+    BIO_set_retry_read(b);
+    return -1;
+  }
+  return len;
+}
+
+int BIO_Puts(BIO* b, const char* str) {
+  return BIO_Write(b, str, strlen(str));
+}
+
+int BIO_Gets(BIO* b, char* buf, int len) {
+  return -1;
+}
+
+long BIO_Ctrl(  // NOLINT(runtime/int)
+    BIO* b,
+    int cmd,
+    long num,  // NOLINT(runtime/int)
+    void* ptr) {
+  return cmd == BIO_CTRL_FLUSH ? 1 : 0;
+}
+
+int BIO_Create(BIO* b) {
+  BIO_set_init(b, 1);
+  return 1;
+}
+
+int BIO_Destroy(BIO* b) {
+  return b == nullptr ? 0 : 1;
+}
+}  // namespace
+
+BIO_METHOD* CreateBIOMethod() {
+  static BIO_METHOD* method = nullptr;
+
+  if (method == nullptr) {
+    method = BIO_meth_new(BIO_TYPE_FD, "bio");
+    BIO_meth_set_write(method, BIO_Write);
+    BIO_meth_set_read(method, BIO_Read);
+    BIO_meth_set_puts(method, BIO_Puts);
+    BIO_meth_set_gets(method, BIO_Gets);
+    BIO_meth_set_ctrl(method, BIO_Ctrl);
+    BIO_meth_set_create(method, BIO_Create);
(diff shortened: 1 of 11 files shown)


jasnell:
I'm stepping back and redoing this PR... will be updating soon /cc @addaleax

jasnell:
ping @addaleax @danbev @trivikr

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1827 @@
+#include "node_quic_crypto.h"
```
jasnell: Review note: previously this was all inlined functions in node_quic_crypto.h but that really never made sense (well, it did very early on but stopped making sense a while ago). Unfortunately, quite a few of the functions have changed but the move here makes that harder to see.

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1827 @@
...
+    BIO_meth_set_puts(method, BIO_Puts);
+    BIO_meth_set_gets(method, BIO_Gets);
+    BIO_meth_set_ctrl(method, BIO_Ctrl);
+    BIO_meth_set_create(method, BIO_Create);
+    BIO_meth_set_destroy(method, BIO_Destroy);
+  }
+  return method;
+}
+
+// TODO(@jasnell): Replace with ngtcp2_crypto_aead_keylen once
```
jasnell: The ngtcp2_crypto helper library is part of ngtcp2 and is, for the most part, straightforward. However, there are a couple of functions in ngtcp2_crypto that depend directly on BoringSSL QUIC APIs that are currently being ported to OpenSSL 3. Those are still a WIP and have a copyright issue that still needs to be resolved before we could even begin to rely on them. So, we will either need to (a) wait to adopt ngtcp2_crypto or (b) we float a patch that replaces the BoringSSL APIs with what we have available today or (c) we attempt to port the BoringSSL APIs into the OpenSSL 1.1.1 we have now. I'm currently assuming option (a)

src/node_quic_session.cc
```diff
@@ -50,7 +50,7 @@ namespace quic {
 // QuicSession is an abstract base class that defines the code used by both
 // server and client sessions.
 QuicSession::QuicSession(
-    QuicSessionType type,
+    ngtcp2_crypto_side side,
```
jasnell: `ngtcp2_crypto_side` is actually defined in the newer version of ngtcp2 that we will be moving to. I'e gone ahead and adopted it here to make moving to the updated ngtcp2 in the next iteration easier.

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1829 @@
...
+  return 1;
+}
+
+void Transport_Params_Free_CB(
+    SSL* ssl,
+    unsigned int ext_type,
+    unsigned int context,
+    const unsigned char* out,
+    void* add_arg) {
+  delete[] const_cast<unsigned char*>(out);
```
addaleax:
```suggestion
  delete[] reinterpret_cast<const uint8_t*>(out);
```

(`delete[]` should be able to work on a const array, but making the types match should make this a bit more “obviously correct”)

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1829 @@
...
+  SetupInitialCryptoContext(&ctx);
+
+  const size_t addrlen = SocketAddress::GetAddressLen(addr);
+  size_t keylen = aead_key_length(&ctx);
+  size_t ivlen = packet_protection_ivlen(&ctx);
+
+  uint64_t now = uv_hrtime();
+
+  auto p = std::begin(plaintext);
+  p = std::copy_n(reinterpret_cast<const uint8_t *>(addr), addrlen, p);
```
addaleax:
```suggestion
  p = std::copy_n(reinterpret_cast<const uint8_t*>(addr), addrlen, p);
```

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1829 @@
...
+  p = std::copy_n(reinterpret_cast<uint8_t *>(&now), sizeof(now), p);
```
addaleax:
```suggestion
  p = std::copy_n(reinterpret_cast<uint8_t*>(&now), sizeof(now), p);
```

src/node_quic_crypto.h
```diff
@@ -1261,549 +313,63 @@ inline int Server_Transport_Params_Parse_CB(
...
-  } else {
-    prefix = "";
-    suffix = sub_parts[0];
-  }
-
-  if (prefix.length() + suffix.length() > host_parts[0].length())
-    return false;
+void GetCertificateAltNames(
+    X509* cert,
+    std::unordered_multimap<std::string, std::string>* map);
```
addaleax: This could just be a return value here, right?

src/node_quic_crypto.cc
```diff
@@ -0,0 +1,1829 @@
...
+    }
+    case SSL3_RT_ALERT: {
+      const uint8_t* msg = reinterpret_cast<const uint8_t*>(buf);
+      CHECK_EQ(len, 2);
+      if (msg[0] == 2)
+        session->SetTLSAlert(msg[1]);
+      break;
+    }
+    default:
+      // Fall-through
```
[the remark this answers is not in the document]
jasnell: Heh, good point ;-)

src/node_quic_crypto.h
jasnell: Yeah, it could.

src/string_bytes.cc
```diff
@@ -607,6 +608,12 @@ size_t StringBytes::hex_encode(
   return dlen;
 }
 
+std::string StringBytes::hex_encode(const char* src, size_t slen) {
+  size_t dlen = slen * 2;
+  char* dst = node::UncheckedMalloc(dlen);
+  hex_encode(src, slen, dst, dlen);
+  return std::string(dst, dlen);
```
addaleax: The `UncheckedMalloc` call leaks memory :)

You should be able to avoid the extra allocation anyway by using something along the lines of 

```c++
  size_t dlen = slen * 2;
  std::string dst(dlen);
  hex_encode(src, slen, &dst[0], dlen);
  return dst;
```

jasnell:
Landed

PR #123, opened 2019-09-17
quic: fix cleanup hooks for Timer class

addaleax:
Just something I noticed while looking at #104.

---

The cleanup hook needs to be removed from the destructor, otherwise
the environment will keep a dangling pointer to the Timer instance
lying around, and the cleanup hook needs to actually close the
libuv handles associated with the Environment’s event loop.


jasnell:
Landed

PR #124, opened 2019-09-18
quic: fix lint issues in node_quic_crypto.cc

danbev:
This commit fixes the following lint issues:
```console
src/node_quic_crypto.cc:136: Use left leaning pointer instead of right leaning
src/node_quic_crypto.cc:147: Use nullptr instead of NULL
src/node_quic_crypto.cc:152:
Use left leaning pointer instead of right leaning
src/node_quic_crypto.cc:162: Use nullptr instead of NULL
src/node_quic_crypto.cc:167:
Use left leaning pointer instead of right leaning
src/node_quic_crypto.cc:176:  Use nullptr instead of NULL
src/node_quic_crypto.cc:247:
Use left leaning pointer instead of right leaning
```

##### Checklist


addaleax:
Fwiw, the `TODO` comments lead me to believe that these functions might be based on copies from ngtcp2_crypto.

If we do fix styling issues, could you also indent the `switch` statement bodies as we usually do?

danbev:
@addaleax I should have read the TODO :( Even though these will be replaced it would be nice to have the linter pass as it is extra noice for me at the moment while looking into locally failing tests.

> If we do fix styling issues, could you also indent the `switch` statement bodies as we usually do?

Absolutely.

jasnell:
Yep, these are temporary but it's difficult to say how temporary they are. It might be some time before we can switch over entirely to ngtcp2_crypto so making sure they lint properly is ++

danbev:
Landed in 925e4befc13ac29d1214580cf89962d97321292c.

PR #126, opened 2019-09-21
C++-side refactor

addaleax:
##### quic: remove `typedef enum`

Using `typedef enum` is a C-ism.

##### quic: remove user_data from quic buffer callback

Using `std::function` and an opaque pointer are pretty much
antithetical. This commit chooses the more C++-y variant, i.e.
`std::function`.

##### quic: cleanup node_quic_buffer.h

Prefer inline functions over macros, use explicit nullptr checks,
do not mark functions with implicit inline linkage as inline,
re-use code where possible, etc.

##### src: cleanup histogram code

Move methods that do not need to be inline into their own file,
replace `std::function` with a template variant for performance,
remove `inline` for functions with implicit inline linkage.

##### quic: remove opaque pointer from Timer callback

Use the more C++-y `std::function` approach which renders opaque
pointers unnecessary.

##### quic: cleanup QuicSession code

- Do not create new `v8::Function` objects, both for performance
  and because it is unclear whether that leaks memory
  (Node.js issue 28988).
- Do not put functions into the inline header that do not need to
  be there or that do not make sense as inline functions
  (in particular, callbacks that are provided to C libraries can
  not be realized as inline functions by the compiler).
- Remove implicit/redundant `inline` qualifiers and add appropriate
  `const` qualifiers.
- Remove unnecessary `this->` prefixes.
- Return early for JS failures rather than accumulating them.
- Minor stylistic changes.

##### src: refactor memory tracker implementation

Use a template class instead of virtual methods to allow inlining
the subclass methods, as well as other minor cleanups.

##### http2: use shared memory tracking implementation

The extensive testing done on http2 makes it easier to make sure
the implementation is correct (and doesn’t diverge unnecessarily).

##### quic: remove QuicStreamListener

This listener currently had the drawbacks of leaking memory
when an error is encountered and was otherwise equivalent to
the default listener.

##### quic: more general C++ cleanup

Use default initializers, `const` qualifiers, minor style changes
to match more common style in the codebase.


src/node_quic_util.h
```diff
@@ -435,16 +431,12 @@ class Timer {
...
-
   static void OnTimeout(uv_timer_t* timer);
   static void CleanupHook(void* data);
 
   bool stopped_;
   Environment* env_;
-  std::function<void(void* data)> fn_;
+  std::function<void()> fn_;
   uv_timer_t timer_;
   void* data_;
```
danbev: Looks like the `data_` member is not used anymore:
```console
In file included from ../src/node_quic_util.cc:1:
../src/node_quic_util.h:440:9: warning: private field 'data_' is not used [-Wunused-private-field]
  void* data_;
        ^
1 warning generated.
```

src/histogram.cc
```diff
@@ -0,0 +1,111 @@
+#include "histogram.h"
+#include "histogram-inl.h"
```
danbev: Should this be only `histogram-inl.h` or perhaps have an lint exclude (`// NOLINT(build/include_inline)`)?

src/node_quic_util.h
addaleax: Yup, done :+1:

src/histogram.cc
addaleax: Yeah, thanks … feels a bit silly that you have to have a linter exception for cases like these, though?

addaleax:
Landed in 72997d659a84...ea095af59f9ff209b5044f5be634460f5b7b35e

PR #127, opened 2019-09-23
quic: make GetAddress take host and host_len

danbev:
Currently, GetAddress takes a char** and assigns to a stack allocated
char buffer. This will cause logging in QuicSocket::SendPacket to look
like this:
```console
QuicSocket (0:3) Sending to  at port 1234
```
This should instead be something like:
```console
QuicSocket (0:3) Sending to 127.0.0.1 at port 1234
```
This commit changes the host parameter of GetAddress to be a char* and,
adds a host_len parameter which should be the length of the char array.
This seems to be more efficient than allocating/copying but I'd be happy
to revisit it if others feel otherwise.

##### Checklist


danbev:
Landed in 42ec7cb5b6531111f12849e3a61d7d7352dca1ed.

Issue #39, opened 2019-06-19
Connection timeout not yet implemented


splitice:
Direction: Client -> Server
Description: Connection timeout not yet implemented

jasnell:
Idle timeouts are implemented at the C++ level and have been implemented for both QuicClientSession and QuicServerSession.

splitice:
Unless my build is derped (5 day old iteration-7) it is non functional, no timeout (or event of any kind) occurs on connection to a blackhole (i.e crashed quic server).

I can repeat testing with merged master once iteration-7 merges.

splitice:
FYI not abandoned. Just haven't had time to do this yet. I plan to do so against draft-22 now that it's merged.

jasnell:
Nice. Ok. From this point on I'm expecting to do much smaller prs in general

splitice:
Unfortunately wasnt able to achieve a connection with this build (same JavaScript code as earlier).

PCAP shows two packets exchanged.

05:21:48.254978 IP *.*.*.*.40407 > *.*.*.*.1443: UDP, length 1252
05:21:48.256958 IP *.*.*.*.1443 > *.*.*.*.40407: UDP, length 71

Connecting code:
```
    static async connect(address, port, client){
        const socket = createSocket({ type: 'udp4', port: 0, idleTimeout:10 });
        const deferred = Q.defer()
        client.logger.info("connecting to %s:%s", address, port)
        const clientSession = socket.connect({address, port, idleTimeout:10})
        clientSession.on('secure',  async () => {
            client.logger.info("secure and ready for action")
            const url = await doAuthenticate(clientSession, client)
            deferred.resolve(new TunnelConnection(clientSession, url, client.timers))
        }).on('error', deferred.reject)
        
        return await deferred.promise
    }
```

Never makes it past "connecting to [...]"

jasnell:
Also, is the `test\parallel\test-quic-client-server` test working for you?

splitice:
Server code is the same as we ran in the it-7 test.

```
    server = createSocket({ type: 'udp4', port: Port });

    // Tell the socket to operate as a server...
    server.listen({ key, cert });

    server
        .on('error', err => debug("server (1) error: %s", err))
        .on('session', (session) => {
//. ...
        })
    deferred.resolve()
```

As I see it this highlights two issues:
1. The client doesnt timeout on this failed connection (e.g by emitted error event)
2. Whatever issue is preventing the initialization of the session / connection.

[...]

PR #128, opened 2019-09-25
quic: use std::shared_ptr for AddStream

addaleax:
This is based on https://github.com/nodejs/quic/issues/39#issuecomment-535014207. Note that I’ll probably work more on this in the near future and that that might actually mean moving away from `std::shared_ptr` and figuring out a better solution (and maybe one that also works well for HTTP/2 where we’ve had similar issues), because currently it’s pretty easy to crash the process during cleanup because ownership of objects is conflictingly both managed through `std::shared_ptr`s and through the association with JS objects.

---

We stored the argument as a shared_ptr in the `streams_` map,
and `QuicStream`s are managed through shared_ptr instances,
so this method should use a shared_ptr argument as well.

This requires moving the `AddStream()` call to until after
the object construction is finished.

##### Checklist

### DIFF
--- a/src/node_quic_session.cc
+++ b/src/node_quic_session.cc
@@ -332,7 +332,7 @@ void QuicSession::AckedStreamDataOffset(
 
 // Add the given QuicStream to this QuicSession's collection of streams. All
 // streams added must be removed before the QuicSession instance is freed.
-void QuicSession::AddStream(QuicStream* stream) {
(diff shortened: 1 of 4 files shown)


jasnell:
I'd definitely be in favor of moving away from shared_ptr

jasnell:
Landed

PR #129, opened 2019-09-25
quic: do not leak memory when UDP sending fails

addaleax:
Destroy `SendWrap` instances that are not destroyed through the
usual path because sending failed.

There should possibly be error handling for some of
those failure conditions, too.

##### Checklist


jasnell:
Landed

PR #130, opened 2019-09-25
quic: manage event loop request counter for `QuicSocket::SendWrap`s

addaleax:
As we don’t use `ReqWrap` like elsewhere in the code base, we need
to manage the counter manually as a way to make sure that the
Environment exists for as long as necessary when shutting down while
a `SendWrap` is in progress.

##### Checklist


jasnell:
Landed

PR #131, opened 2019-09-25
quic: remove memcmp before memcpy

addaleax:
This pattern is equivalent to just performing memcpy.
The previous code would be reported as access to uninitialized
memory by e.g. valgrind.

##### Checklist


jasnell:
Landed

PR #133, opened 2019-09-25
quic: fix native classes to work well during teardown

addaleax:
##### quic: make allocators first members of Quic classes
    
This makes sure that the allocators outlive all other members,
and in particular are available for ngtcp2 cleanup calls.

##### quic: adjust expectations in destructors
    
Some expectations that were part of the destructors may not always
be met; in particular, when no explicit destruction of the objects
had happened before.

##### Checklist


src/node_quic_session.h
```diff
@@ -876,8 +878,6 @@ class QuicSession : public AsyncWrap,
 
   AliasedFloat64Array state_;
 
-  ngtcp2_mem alloc_info_;
```
addaleax: Fwiw, I think these re-ordering changes shouldn’t affect anything in practice, but I *think* keeping these where they are could otherwise result in undefined behaviour. This is basically just making sure.

jasnell: Makes perfect sense

jasnell:
Landed!

PR #135, opened 2019-09-26
quic: always set the first byte of sockaddr zero

oyyd:
The tests related to quic time out on my mac. After some investigation, I find that the first byte of `sockaddr` might not be zero on FreeBSD and macOS. Sometimes they are [`sa_len`](https://www.freebsd.org/doc/en_US.ISO8859-1/books/developers-handbook/sockets-essential-functions.html) which depends on how we get `sockaddr` from uv. 

And `ngtcp2` might drop those packets of which `sockaddr`s don't match by bytes: https://github.com/ngtcp2/ngtcp2/blob/d61344c6a3441ebd60e01438bfa35052e457d8f1/lib/ngtcp2_conn.c#L6607

After this modification, these tests work fine on my mac.

##### Checklist


addaleax:
Also, thanks for investigating into why this didn’t work on MacOS! :blue_heart:

oyyd:
@addaleax Right. It seems to be an issue of libuv on Mac:

```c
#include <netinet/ip.h>
#include <assert.h>
#include <stdio.h>
#include <uv.h>

static void print_sockaddr(const struct sockaddr *addr) {
  uint8_t* addr_arr = (uint8_t*)(addr);

  for (size_t i = 0; i < sizeof(struct sockaddr); i += 1) {
    fprintf(stderr, "%d ", addr_arr[i]);
  }
  fprintf(stderr, "\n");
}

int main() {
  uv_loop_t loop;
  uv_loop_init(&loop);

  uv_udp_t udp;
  uv_udp_init(&loop, &udp);

  struct sockaddr_in addr;
  uv_ip4_addr("127.0.0.1", 0, &addr);
  // print something like: 0 2 0 0 127 0 0 1 0 0 0 0 0 0 0 0
  print_sockaddr(&addr);

  assert(!uv_udp_bind(&udp, (const struct sockaddr*)(&addr), 0));

  struct sockaddr_storage addr_bind;
  int namelen = sizeof(struct sockaddr_in);
  // uv_udp_getsockname() will call getsockaddr()
  assert(!uv_udp_getsockname(&udp, (struct sockaddr*)(&addr_bind), &namelen));

  // print something like: 16 2 198 138 127 0 0 1 0 0 0 0 0 0 0 0
  print_sockaddr(&addr_bind);

  // uv_run(&loop, UV_RUN_DEFAULT);

  return 0;
}
```

`uv_ip4_addr` and `uv_ip6_addr`  directly manipulate `struct sockaddr_in`: https://github.com/libuv/libuv/blob/v1.x/src/uv-common.c#L210 when `uv_udp_getsockname` finally calls `getsockname`. 

I guess a more appropriate solution is to make libuv set `sa_len` inside of `uv_ip4_addr` and `uv_ip6_addr`.

addaleax:
> I guess a more appropriate solution is to make libuv set `sa_len` inside of `uv_ip4_addr` and `uv_ip6_addr`.

Yeah, I think that would make sense. There is already corresponding code for IPv6, but it’s behind a `#ifdef SIN6_LEN` check and I don’t think that’s compiled with Node.js either.

I’m wondering if it’s really ideal to pass the whole `sockaddr` struct to ngtcp2, though… Like, this is an issue that could somewhat easily be circumvented by using a custom struct rather than the platform-specific `sockaddr` (that could also have additional padding, too).

jasnell:
I'd *really* like a custom struct from libuv on this, to be honest. Using sockaddr directly is extremely cumbersome and, as evidenced, can be problematic across platforms. I've never really liked that libuv made us work directly with sockaddr.

That's not a trivial bit of work to do on the libuv side, however, and could take a while so fixing this here is 💯... thank you so much for investigating.

jasnell:
just a heads up, this will need to be rebased against master

oyyd:
https://github.com/libuv/libuv/pull/2492 also solves this issue so that I'm going to close this.

PR #138, opened 2019-09-27
quic: update ngtcp2, openssl APIs, and use ngtcp2-crypto

jasnell:
Adopts the BoringSSL APIs based on https://github.com/openssl/openssl/pull/8797. We may need to update the license accordingly. These will need to be tracked over time.

Update the implementation to use the new ngtcp2_crypto helper.

Note: because of the need to use the BoringSSL APIs, QUIC will not be available when using shared_openssl. Still need to make sure that's disabled in the build.

##### Checklist


jasnell:
@addaleax ... these are the WIP updates I've been working through. They are incomplete and I haven't yet checked to see if all of the build issues are resolved. I'll be doing that next, but this should be enough to start digging through if you want to take a look (no worries if you don't, just wanted to give you a heads up :-) ...)

jasnell:
There are a number of known issues with this so far. I just started the changes over again after rebasing the latest round of changes. test-quic-client-server fails immediately. I'll be looking at that next but may not get to it today.

jasnell:
Quick update... Ok, got this back to the point where it's building successfully and running up to the point where the handshake gets stalled (where it was previously before I redid the whole thing). Investigating that failure next.

jasnell:
Btw, here's what I've narrowed down so far...

```
QuicSocket (0:13) Receiving 1252 bytes from the UDP socket.
QuicSocket (0:13) Received a QUIC packet for dcid ba69759ba401686a02b7577a82c8f5a376692760
QuicSession Client (17) Receiving QUIC packet.
QuicSession Client (17) Processing received packet
QuicSession Client (17) Receiving 123 bytes of crypto data.
QuicSession Client (17) Writing 123 bytes of peer handshake data.
QuicSession Client (17) TLS handshake continuing
QuicSession Client (17) Buffered peer handshake bytes: 123, offset: 0
QuicSession Client (17) Reading 123 bytes of peer handshake data starting at offset 0
QuicSession Client (17) Buffered peer handshake bytes: 123, offset: 123
QuicSession Client (17) Reading 0 bytes of peer handshake data starting at offset 123
```

Notice that the client QuicSocket receives 1252 bytes from the server, which is then passed off to the QuicClientSession for processing. ngtcp2 is given all 1252 bytes but only processes the first 123, then immediately buffers and defers processing the rest and I'm not exactly sure why yet.

jasnell:
Ok, I've narrowed it down. ngtcp2 is buffering the handshake data and exiting because the handshake keys on the client side are not installed at the right time. ngtcp2 made a change that assumes that both the rx and tx handshake keys are available at the same time, which is based on the assumption that the boringssl quic APIs are being used. The timing of the KeyCB approach we currently have is different so we're running into an issue. Very happy that I finally narrowed that down but now need to find a fix for it...

jasnell:
Ah, well, this is going to be a fun one... Digging into it... tatsuhiro-san's initial modification to openssl to support quic included adding the key callback we use the get the keying material. That version of the quic modifications would invoke the key callback once per generated secret by patching into openssl's existing `tls13_change_cipher_state` function. The BoringSSL api approach is to actually branch to an entirely different `quic_change_cipher_state` function that generates the client and server secrets for each cipher level at the same time. This is a cleaner and more correct approach but to get it working for us since we don't have the boringssl APIs yet, we're going to have to patch on top of the original patch. 🎉 💯 The moral of the story is that as soon as we can move to the boringssl APIs, the better. I'll start working up the additional openssl patch tomorrow

jasnell:
will be rebasing this against master soon

jasnell:
Ok, progress! With the two additional commits here we get a lot further along in the TLS handshake but we're still not all the way yet. Still digging...

image

deps/openssl/openssl/ssl/ssl_locl.h
```diff
@@ -1125,9 +1125,12 @@ struct ssl_st {
...
     void *msg_callback_arg;
-    int (*key_callback)(SSL *ssl, int name, const unsigned char *secret,
-                        size_t secretlen, void *arg);
-    void *key_callback_arg;
+    int (*encryption_secrets_callback)(SSL* ssl, int level,
+                                       const uint8_t *read_secret,
+                                       const uint8_t *write_secret,
+                                       size_t secret_len,
+                                       void *arg);
+    void *encryption_secrets_callback_arg;
```
jasnell: btw, I *think* these changes are actually ABI breaking in openssl.

addaleax: The changes to `deps/openssl/openssl/include` are ABI- and API-breaking, yes … if you want compatibility, I think you would ideally want to retain `key_callback` (and its setter), and fall back to using it when no `encryption_secrets_callback` is set, if that’s possible

jasnell: key_callback is not part of the OpenSSL API. tatsuhiro-san added those in his initial hack to get QUIC support working. This is replacing his non-standard api with another non-standard api

addaleax: I think we’re good then – C function additions generally are ABI- and API-stable, and anything that doesn’t touch the include folder is okay, too.

jasnell:
Testing is confirming that we are now completing the client side of the TLS handshake! The server side, however, is still not completing. I'll track that down tomorrow.

jasnell:
PROGRESS! Ok, with a couple of exceptions, things should be working. There's still work to do in this PR to get things cleaned up. The one thing that's not working is accessing the client certificate on the server side. That's what I'll be chasing down tomorrow. But, progress has been made.

jasnell:
WOO! Ok, this should be ready to go. There are still a few pieces that need to be revisited but we need some fixes in ngtcp2 to resolve thos (see https://github.com/ngtcp2/ngtcp2/issues/156)

jasnell:
There's one commit in here that needs to be upstreamed... https://github.com/ngtcp2/ngtcp2/pull/157

src/node_quic_crypto.cc
```diff
@@ -1804,5 +805,121 @@ Local<Value> GetCipherVersion(Environment* env, SSL* ssl) {
...
+  SSL_set_alpn_protos(
+      ssl,
+      reinterpret_cast<const uint8_t*>(alpn.c_str()),
+      alpn.length());
+}
+
+void SetHostname(SSL* ssl, const std::string& hostname) {
+  if (hostname.length() == 0 ||
+      SocketAddress::numeric_host(hostname.c_str())) {
+    SSL_set_tlsext_host_name(ssl, "localhost");
```
addaleax: Is it intentional that all IP addresses are implicitly set to `localhost`?

src/node_quic_socket.cc
```diff
@@ -656,7 +658,9 @@ BaseObjectPtr<QuicSession> QuicSocket::AcceptInitialPacket(
   switch (QuicServerSession::Accept(&hd, data, nread)) {
     case QuicServerSession::InitialPacketResult::PACKET_VERSION:
       SendVersionNegotiation(version, dcid, scid, addr);
-      // Fall-through to ignore packet
+      // Return explicitly here rather than falling-through to
+      // avoid compiler warning
+      return {};
```
addaleax: I think we just need to use a fall through comment format that the compiler understands, `// Fall through` or something might work :)

src/node_quic_crypto.h
```diff
@@ -260,62 +97,11 @@ int ALPN_Select_Proto_CB(
...
-    size_t inlen,
-    X509* x,
-    size_t chainidx,
-    int* al,
-    void* parse_arg);
-
 bool GenerateRetryToken(
     uint8_t* token,
-    size_t* tokenlen,
+    size_t& tokenlen,
```
addaleax: Does the linter allow this? We generally have a policy against non-const references, and I think this is a case where using a pointer made sense

src/node_quic_socket.cc
```diff
@@ -555,7 +555,7 @@ void QuicSocket::SendVersionNegotiation(
   if (req->Send() != 0) delete req;  // TODO(addaleax): Better error handling?
 }
 
-ssize_t QuicSocket::SendRetry(
+bool QuicSocket::SendRetry(
```
addaleax: Is this change being used somewhere? It conflicts with what I’m currently doing :)

addaleax: Like, I can mirror these changes, it’s not an issue, it just would be good to know if there’s more to it than I’m seeing, because as of right now the return value is unused afaict

src/node_quic_crypto.cc
jasnell: Ah, that's temporary, just forgot to add a `TODO` for that.

src/node_quic_socket.cc
jasnell: Simplifying and making things consistent. Generally, I'm trying to get rid of as many of the ssize_t uses as possible to simplify error handling.

jasnell: But in general, we may be able to just make this a void return because we're not currently using the return value at all

jasnell:
one more set of minor updates coming in... then will be done..

jasnell:
Landed

PR #141, opened 2019-09-29
src,quic: add custom smart pointer for BaseObjects

addaleax:
This is in response to conversation with @jasnell; the previous approach taken in the QUIC codebase, relying mostly on `std::shared_ptr` and `std::unique_ptr`, conflicts with `BaseObject`’s own lifetime management. In order to avoid these conflicts, this introduces a new type of smart pointer to Node.js specifically for managing and observing the lifetime of `BaseObject` subclass instances.

The first two commits are from node/master, included so that semantic conflicts with them can be avoided early. These will go away once we rebase.

What also came out of that conversation is that, whether we take this approach or not, some documentation on the `BaseObject` lifetime management (and the `Environment` cleanup functionality) is necessary. I don’t think it should hold up this PR, but it would make a good follow-up, especially given that this PR does modify the semantics there and such documentation would look different depending on whether this PR lands.

---

Main commit:

#####  src: introduce custom smart pointers for `BaseObject`s

Referring to `BaseObject` instances using standard C++ smart pointers
can interfere with BaseObject’s own cleanup mechanisms
(explicit delete, delete-on-GC and delete-on-cleanup).

Introducing custom smart pointers allows referring to `BaseObject`s
safely while keeping those mechanisms intact.

##### Checklist


src/base_object.h
```diff
@@ -103,12 +110,44 @@ class BaseObject : public MemoryRetainer {
...
+  // This creates a PointerData struct if none was associated with this
+  // BaseObject before.
+  inline PointerData* pointer_data();
+
+  // Functions that adjust the strong pointer count.
+  inline void decrease_refcount();
+  inline void increase_refcount();
+
   Environment* env_;
+  PointerData* pointer_data_ = nullptr;
```
addaleax: Btw, I’ve considered merging the two members into one tagged pointer (`Environment*` by default, `PointerData*` otherwise + an `Environment*` member in `PointerData`) in order to avoid the extra memory overhead for most BaseObjects, but playing around with similar code and looking at the compiler output I got worried about the added CPU cost, and so I let that be for now.

jasnell:
@addaleax ... with regards to the two upstream commits, should I go ahead and land this now or pull in the latest round of updates from nodejs/node then drop the extra commits from this PR?

addaleax:
@jasnell I’ve pushed another commit with the move constructors suggested in the `TODO` comment and more tests. (Just so you’re aware as the sole reviewer so far.)

> @addaleax ... with regards to the two upstream commits, should I go ahead and land this now or pull in the latest round of updates from nodejs/node then drop the extra commits from this PR?

I don’t think it makes much of a difference – the commits should be dropped automatically once we rebase. Feel free to go with what’s easier for you :)

jasnell:
Landed!

PR #144, opened 2019-10-01
quic: refactor transport_params usage

addaleax:
- Use `Buffer::Copy()` to avoid use-after-free crashes.
- Store the struct directly to save memory, as opposed to using
  the (by default larger) `MaybeStackBuffer`.

It is somewhat unclear to me whether the field will be used after
emitting it through the `sessionTicket`, e.g. through multiple
`.setSession()` calls. If not, this could be further reduced to
an `AllocatedBuffer` from which we create the Buffer instance.

##### Checklist


jasnell:
Nice. I *think* these are only going to be needed up until the `sessionTicket` event is emitted, and even then, only if that event actually has a handler. I was asking the same questions back when I originally added this and decided just to leave it as a placeholder for now.

PR #149, opened 2019-10-03
src,lib,test: make tests mostly pass

addaleax:
Make the test mostly pass, with and without `./configure --experimental-quic`. This does not fix a code cache test, which seems to be broken with and without the flag.


jasnell:
Landed!

PR #150, opened 2019-10-03
quic: implement sendFD() support

addaleax:
Fun fact: The first commit was written in March 2018, when I started working on something cool but never got anywhere with it … finally something to use it for :)

This is mostly based on the HTTP/2 implementation for the same thing.

TODO:

- [x] Docs
- [x] Tests for error conditions
- [x] Figure out why `.close()` is not enough to shut down the test
- [x] `sendFile()` with a file path
- [x] Direct support for `FileHandle` as returned from `fs.promises.open` (this goes for HTTP/2 too)


test/parallel/test-quic-send-fd.js
```diff
@@ -0,0 +1,70 @@
...
+
+  req.on('stream', common.mustCall((stream) => {
+    const data = [];
+    stream.on('data', (chunk) => data.push(chunk));
+    stream.on('end', common.mustCall(() => {
+      assert.deepStrictEqual(Buffer.concat(data), fs.readFileSync(__filename));
+
+      // TODO(addaleax): Figure out why .close() is insufficient.
+      client.destroy();
+      server.destroy();
```
jasnell: Try moving the `client.close()` and `server.close()` to the `on('close')` event and see if that helps. On the end event, we can't be certain that all of the data has been transferred successfully (there still might be retransmissions, etc). The 'close' event tells us that the underlying `QuicStream` has fully completed. tl;dr destroying in end may prematurely cut off data transmission

jasnell:
Oh very nice!

test/parallel/test-quic-send-fd.js
```diff
@@ -0,0 +1,70 @@
...
+  ca,
+  rejectUnauthorized: false,
+  maxCryptoBuffer: 4096,
+  alpn: 'meow'
+});
+
+server.on('session', common.mustCall((session) => {
+  session.on('secure', common.mustCall((servername, alpn, cipher) => {
+    const stream = session.openStream({ halfOpen: false });
+    stream.sendFD(fs.openSync(__filename, 'r'));
```
jasnell: Just been thinking about this for a while... wondering if the `sendFD()` shouldn't have a callback or is the `on('close')` event good enough?

Also, should have a variant that tests error conditions and ensures that the `on('error')` event is emitted properly.

test/parallel/test-quic-send-fd.js
```diff
@@ -0,0 +1,70 @@
...
+  common.skip('missing quic');
+
+const assert = require('assert');
+const quic = require('quic');
+const fs = require('fs');
+
+const fixtures = require('../common/fixtures');
+const key = fixtures.readKey('agent1-key.pem', 'binary');
+const cert = fixtures.readKey('agent1-cert.pem', 'binary');
+const ca = fixtures.readKey('ca1-cert.pem', 'binary');
```
jasnell: todo for later... it's so common for us to read all three of these together we can simplify things a bit by adding an additional fixture method... e.g.

something like...
```js
const { key, cert, ca } = fixtures.readTLSKey('agent1-key.pem',
											  'agent1-cert.pem',
											  'ca1-cert.pem');
```

test/parallel/test-quic-send-fd.js
addaleax: @jasnell So … that works, but I would have expected `.close()` to implicitly end all streams that haven’t been `.end()`ed yet.

> tl;dr destroying in end may prematurely cut off data transmission

There is no data transmission in this test that would be relevant here, though. The client doesn’t send anything and the server is done sending at this point.

test/parallel/test-quic-send-fd.js
addaleax: > Just been thinking about this for a while... wondering if the `sendFD()` shouldn't have a callback or is the `on('close')` event good enough?

I’d say `on('close')` is good enough.

> Also, should have a variant that tests error conditions and ensures that the `on('error')` event is emitted properly.

It’s in the PR TODO list :)

test/parallel/test-quic-send-fd.js
jasnell: Yep, that's something we'll have to look at.

For `QuicSession`, `close()` initiates a graceful close, which is supposed to allow all currently open `QuicStream` instances to close and destroy before closing and destroying the `QuicSession`. Calling `close()` should prevent any new `QuicStream` instances from being created or accepted, but it does not tear down the `QuicSession` immediately so long as there is still an open `QuicStream`. So if having the `close()` here doesn't work, we need to revisit. 

Also, it's important that the internal `ngtcp2_conn` is not destroyed while inside one of the ngtcp2 callbacks.

addaleax: > So if having the `close()` here doesn't work, we need to revisit.

It does work, we just have to call `stream.end()` explicitly along with it.

(I’ll push that later together with other changes.)

addaleax:
HTTP/2 PR for FileHandle support: https://github.com/nodejs/node/pull/29876

I’d implement that here depending on whether it lands in HTTP/2.

test/parallel/test-quic-send-fd.js
jasnell: ah... ok that makes sense then. When using `sendFD()` the writable side of the JavaScript stream should be closed immediately without actually closing the stream. That should be straightforward to do.

addaleax: This is on the other side, not the one that uses `sendFD()`… Closing the writable side of the stream is already happening here

Would it be a bad idea to make `.close()` imply `.end()` if the stream has not been ended yet?

jasnell: that *should* be ok.

jasnell:
This will likely need a rebase

addaleax:
Rebased. This PR is ready now, and contains two more upstream commits that are here to prevent conflicts at the next rebase.

doc/api/quic.md
```diff
@@ -1106,6 +1106,54 @@ added: REPLACEME
...
+* `path` {string|Buffer|URL}
+* `options` {Object}
+  * `onError` {Function} Callback function invoked in the case of an
+    error before send.
+  * `offset` {number} The offset position at which to begin reading.
+    Default: `-1`.
+  * `length` {number} The amount of data from the fd to send.
+    Default: `-1`.
+
+Instead of using `quicstream` as a writable stream, send data from a given file
```
jasnell:
```suggestion
Instead of using a `QuicStream` as a writable stream, send data from a given file
```

doc/api/quic.md
```diff
@@ -1106,6 +1106,54 @@ added: REPLACEME
...
+-->
+
+* `fd` {number|FileHandle} A readable file descriptor.
+* `options` {Object}
+  * `offset` {number} The offset position at which to begin reading.
+    Default: `-1`.
+  * `length` {number} The amount of data from the fd to send.
+    Default: `-1`.
+
+Instead of using `quicstream` as a writable stream, send data from a given file
```
jasnell:
```suggestion
Instead of using a `Quicstream` as a writable stream, send data from a given file
```

addaleax:
Landed in 6b97000f220f...a19682ef2835

PR #152, opened 2019-10-05
quic: default options parameter in constructor

trivikr:
We need not default options parameter as `{}` is passed by default in createSocket
https://github.com/nodejs/quic/blob/40e5be0a01a92d85b017f80e4ed61b9b92a1ea47/lib/internal/quic/core.js#L2311-L2315

Moving this change in the function signature.

##### Checklist


trivikr:
Landed in 92c25a6da2dcc580a205c6da832a04c780de8267

PR #139, opened 2019-09-29
quic: fix the misuse of uv_inet_ntop

oyyd:
The return value of `uv_inet_ntop` means success when it's zero.

##### Checklist

- [x] `make -j4 test` (UNIX), or `vcbuild test` (Windows) passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

PR #155, opened 2019-10-07
quic: client should work when receives invalid preferred address

oyyd:
If we set `preferredAddressPolicy` to `accept` on `QuicClientSession`
but receives an invalid preferred address, e.g. we don't define
`preferredAddress` on the server, the `QuicClientSession` should still
work. Before this PR, the included test will cause segmetation fault.

the example in `ngtcp2` always return 0 on `select_preferred_address `: https://github.com/ngtcp2/ngtcp2/blob/52d56a90a7766c71c68362dc85f5fdc866760f47/examples/client.cc#L673

##### Checklist

### DIFF
--- a/src/node_quic_util.h
+++ b/src/node_quic_util.h
@@ -200,7 +200,7 @@ class SocketAddress {
     }
 
     char host[NI_MAXHOST];
-    if (uv_inet_ntop(af, binaddr, host, sizeof(host)) == 0)
+    if (uv_inet_ntop(af, binaddr, host, sizeof(host)) != 0)
       return false;
 
     addrinfo hints{};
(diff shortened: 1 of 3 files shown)


src/node_quic_util.h
```diff
@@ -200,7 +200,7 @@ class SocketAddress {
     }
 
     char host[NI_MAXHOST];
-    if (uv_inet_ntop(af, binaddr, host, sizeof(host)) == 0)
+    if (uv_inet_ntop(af, binaddr, host, sizeof(host)) != 0)
```
oyyd: This seems to be missed from the last rebasing. Refs: https://github.com/nodejs/quic/pull/139/files

jasnell: how odd... thanks for catching

jasnell:
Landed

PR #157, opened 2019-10-08
quic: compute retransmission timeout correctly

addaleax:
Previously, if the expiry timestamp was in the future by less
than 1ms, the retransmission timer would not have been scheduled.

(This also removes an unused line from the test that this
problem made flaky.)

##### Checklist


addaleax:
Let’s actually try to run CI on this…

https://ci.nodejs.org/job/node-test-pull-request/25909/

addaleax:
Landed in 70a65d3b7c0e34f79085480341e3381455807794

PR #158, opened 2019-10-08
deps: cherry-pick udp libuv patches

addaleax:
In particular, this includes @oyyd’s patch to make more tests pass on a lot of platforms. There commits can be dropped at the next rebase.


jasnell:
Landed

PR #160, opened 2019-10-09
test: add tests for quic `idleTimeout`

oyyd:
It would be better to add a test to ensure that the timeout is greater than the triple PTO if we can get the PTO.

##### Checklist

(https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)


jasnell:
Landed

addaleax:
This test is timing out 100 % of the time for me.

PR #162, opened 2019-10-09
quic: fix build and tests

addaleax:
The recent commits have broken the build, tests, and linting.

One cctest is not fixed but rather disabled here.

---

I intend to merge this as soon as it’s open and just let this PR be a notification that this happened, given that we don’t follow the strict merging rules of the main repo.


addaleax:
Landed in 968a23c05f35c519a388e43b77bc8c6a080ae7d1

Issue #80, opened 2019-08-21
TODO: Verify streams opened out of order


jasnell:
QuicStream ... Implement test verifying that "A stream ID that is used out of order results in all streams of that type with lower-numbered stream IDs also being opened."

Type: test
Priority: high

jasnell:
Yeah, this is only for received streams. We would have to find an
artificial or external way of testing this

jasnell:
Have verified that ngtcp2 is handling this for us. There's nothing for us to do here.

PR #165, opened 2019-10-09
quic: share UDP code with dgram

addaleax:
Make `UDPWrap` useable by other C++ classes, and use them for QUIC. This simplifies the code a bit, and will additionally enable us to test QUIC without actual networking usage (e.g. when reproducibility for a test is particularly important).

One commit is coming from upstream Node.js. Once everything is working, I’d expect this to increase performance a bit, too.

TODO: 

- [x] Make all tests pass (currently a number of them time out)
- [x] Internal documentation for the UDPWrapBase / UDPListener interfaces
- [ ] Test QUIC without actual network socket


addaleax:
One of the testing possibilities I’d have in mind is #80 – if that’s something that depends on the order of packets arriving, then let’s control the order in which they arrive to test it :)

jasnell:
Looking good so far!

addaleax:
This is ready for review now, the testing parts should be able to follow later.

src/udp_wrap.h
```diff
@@ -64,6 +146,25 @@ class UDPWrap: public HandleWrap {
   static void SetTTL(const v8::FunctionCallbackInfo<v8::Value>& args);
   static void BufferSize(const v8::FunctionCallbackInfo<v8::Value>& args);
 
+  // UDPListener implementation
+  uv_buf_t OnAlloc(size_t suggested_size) override;
+  void OnRecv(ssize_t nread,
+              const uv_buf_t& buf,
+              const sockaddr* addr,
+              unsigned int flags);
```
danbev: I think this function should have an `override` specifier.

src/js_udp_wrap.cc
```diff
@@ -0,0 +1,223 @@
...
+}
+
+void JSUDPWrap::OnSendDone(const FunctionCallbackInfo<Value>& args) {
+  JSUDPWrap* wrap;
+  ASSIGN_OR_RETURN_UNWRAP(&wrap, args.Holder());
+
+  CHECK(args[0]->IsObject());
+  CHECK(args[1]->IsInt32());
+  ReqWrap<uv_udp_send_t>* req_wrap;
+  ASSIGN_OR_RETURN_UNWRAP(&wrap, args[0].As<Object>());
```
danbev: Should this be `&req_wrap` instead of `&wrap`?

src/udp_wrap.h
jasnell:
```suggestion
              unsigned int flags) override;
```

src/js_udp_wrap.cc
jasnell:
```suggestion
  ASSIGN_OR_RETURN_UNWRAP(&req_wrap, args[0].As<Object>());
```

addaleax:
Sorry it took so long, but – Finally, it works :tada: 

@danbev @jasnell Could you take another look? Here’s what’s changed since your reviews:

- 9cdf8d9c2dc3 quic: allow `.connect()` when socket is already bound
  - This is necessary for the test to work, because the socket can be considered bound immediately.
    Afaict, this should also allow calling `.connect()` on the client side multiple times, with multiple peers, and getting different sessions for them – that would be allowed and good, right?
- 5453076fcb2f squash! quic: allow testing QUIC without real UDP handle
  - This adds an actual in-memory-QUIC test without any real I/O that can be used for more robust tests when the need arises.
- 33951a64b2e6 fixup! dgram: make UDPWrap more reusable
  - [This is just addressing a review comment].
- db35d5ef6122 fixup! quic: use UDP code from dgram
  - Adjusts @oyyd’s test from #171 to account for the better error message that now includes the syscall name.

jasnell:
> Afaict, this should also allow calling .connect() on the client side multiple times, with multiple peers, and getting different sessions for them – that would be allowed and good, right?

I think that's definitely fine.

> This adds an actual in-memory-QUIC test without any real I/O that can be used for more robust tests when the need arises.

🎉 💯 🥇

addaleax:
Landed in 1ff05f49480a...a76f40a810dc

Issue #164, opened 2019-10-09
TODO: fix test-quic-idle-timeout


addaleax:
The test was recently added in #160 but is timing out 100 % of the time for me (x64 Linux).

jasnell:
hmm... I ran it once and it worked, now it's not. Weird.

oyyd:
After pulling the master today, it doesn't work for me too. Need to find out the reason.

addaleax:
Fixed in #166

PR #166, opened 2019-10-10
quic: fix idle timeout

oyyd:
After ngtcp2 being updated, `ngtcp2_conn_get_idle_timeout` is renamed as
`ngtcp2_conn_get_idle_expiry` which returns `ngtcp2_tstamp` instead of
`ngtcp2_duration`.

Refs: https://github.com/ngtcp2/ngtcp2/blob/6f40668cdce7db7c043d3a80c07f379841d8c51e/lib/ngtcp2_conn.c#L8604

Fixes: #164 

##### Checklist

### DIFF
--- a/src/node_quic_session.cc
+++ b/src/node_quic_session.cc
@@ -1527,7 +1527,10 @@ void QuicSession::StreamReset(
 
 void QuicSession::UpdateIdleTimer() {
   CHECK_NOT_NULL(idle_);
-  uint64_t timeout = ngtcp2_conn_get_idle_expiry(Connection()) / 1000000UL;
+  uint64_t now = uv_hrtime();
+  uint64_t expiry = ngtcp2_conn_get_idle_expiry(Connection());
+  uint64_t timeout = (expiry - now) / 1000000UL;
+  if (expiry < now || timeout == 0) timeout = 1;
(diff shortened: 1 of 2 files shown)


src/node_quic_session.cc
```diff
@@ -1527,7 +1527,9 @@ void QuicSession::StreamReset(
 
 void QuicSession::UpdateIdleTimer() {
   CHECK_NOT_NULL(idle_);
-  uint64_t timeout = ngtcp2_conn_get_idle_expiry(Connection()) / 1000000UL;
+  uint64_t timeout = (ngtcp2_conn_get_idle_expiry(Connection()) - uv_hrtime()) / 1000000UL;
+  if (timeout < 0)
+    timeout = 0;
```
addaleax: `timeout < 0` never happens because it is unsigned; and `0` won’t schedule the time afaik, whereas `1` will…

Should this maybe overall look similar to the retransmit timer code as in e.g. https://github.com/nodejs/quic/pull/157/files?

oyyd: That's right.. I would like to follow the https://github.com/nodejs/quic/pull/157/files

addaleax:
Landed in 6b97000f220f2eef87a594969c60bd5c900bb2af, thanks! :)

PR #167, opened 2019-10-10
quic: simplify alpn selection callback

jasnell:
Refinements to alpn selection. Next step will be to error if the alpn selection was not successful


jasnell:
Landed!

PR #168, opened 2019-10-11
quic: simplify QuicSession and isolate handshake logic

jasnell:
This is completing a refactor I had started a while ago before the handshake broke. It greatly simplifies the internal structure by eliminating QuicClientSession and QuicServerSession in favor of just a single QuicSession, and it separates out the crypto functionality into a separate QuicCryptoContext. This should make it significantly easier to reason about internally. The next step is to start rolling in the http3 specific functionality.

### DIFF
--- a/src/node_quic_crypto.cc
+++ b/src/node_quic_crypto.cc
@@ -126,7 +126,7 @@ bool GenerateRandData(uint8_t* buf, size_t len) {
 Local<Array> GetClientHelloCiphers(QuicSession* session) {
   const unsigned char* buf;
   Environment* env = session->env();
-  auto ctx = session->CryptoContext();
+  QuicCryptoContext* ctx = session->CryptoContext();
   size_t len = SSL_client_hello_get0_ciphers(**ctx, &buf);
   std::vector<Local<Value>> ciphers_array;
   for (size_t n = 0; n < len; n += 2) {
@@ -135,14 +135,14 @@ Local<Array> GetClientHelloCiphers(QuicSession* session) {
     const char* cipher_name = SSL_CIPHER_get_name(cipher);
     const char* cipher_version = SSL_CIPHER_get_version(cipher);
     Local<Object> obj = Object::New(env->isolate());
-    USE(obj->Set(
+    obj->Set(
         env->context(),
         env->name_string(),
-        OneByteString(env->isolate(), cipher_name)));
-    USE(obj->Set(
+        OneByteString(env->isolate(), cipher_name)).FromJust();
+    obj->Set(
         env->context(),
         env->version_string(),
-        OneByteString(env->isolate(), cipher_version)));
+        OneByteString(env->isolate(), cipher_version)).FromJust();
     ciphers_array.push_back(obj);
   }
   return Array::New(env->isolate(), ciphers_array.data(), ciphers_array.size());
@@ -153,7 +153,7 @@ const char* GetClientHelloServerName(QuicSession* session) {
     size_t len;
     size_t rem;
 
-    auto ctx = session->CryptoContext();
+    QuicCryptoContext* ctx = session->CryptoContext();
 
     if (!SSL_client_hello_get0_ext(
             **ctx,
@@ -187,7 +187,7 @@ const char* GetClientHelloALPN(QuicSession* session) {
     size_t len;
     size_t rem;
 
-    auto ctx = session->CryptoContext();
+    QuicCryptoContext* ctx = session->CryptoContext();
 
     if (!SSL_client_hello_get0_ext(
             **ctx,
@@ -707,7 +707,7 @@ int VerifyHostnameIdentity(SSL* ssl, const char* hostname) {
 }
 
 const char* GetServerName(QuicSession* session) {
-  auto ctx = session->CryptoContext();
+  QuicCryptoContext* ctx = session->CryptoContext();
   return SSL_get_servername(**ctx, TLSEXT_NAMETYPE_host_name);
 }
 
@@ -731,7 +731,7 @@ Local<Value> GetALPNProtocol(QuicSession* session) {
   Local<Value> alpn;
   const unsigned char* alpn_buf = nullptr;
   unsigned int alpnlen;
-  auto ctx = session->CryptoContext();
+  QuicCryptoContext* ctx = session->CryptoContext();
 
   SSL_get0_alpn_selected(**ctx, &alpn_buf, &alpnlen);
   if (alpnlen == sizeof(NGTCP2_ALPN_H3) - 2 &&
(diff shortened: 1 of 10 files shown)


src/node_quic_crypto.cc
```diff
@@ -713,56 +727,128 @@ Local<Value> GetServerName(
...
+    value = crypto::X509ToObject(session->env(), cert);
+  return value;
+}
+
+Local<Value> GetEphemeralKey(QuicSession* session) {
+  Environment* env = session->env();
+  Local<Context> context = env->context();
+
+  Local<Object> info = Object::New(env->isolate());
+  auto ctx = session->CryptoContext();
```
addaleax: You don’t have to agree with me, but I still kind of don’t like `auto` here … I’d still always have to look up the type if I wanted to know it.

src/node_quic_crypto.cc
```diff
@@ -713,56 +727,128 @@ Local<Value> GetServerName(
...
+            EC_KEY_free(ec);
+          } else {
+            curve_name = OBJ_nid2sn(kid);
+          }
+          USE(info->Set(context, env->type_string(),
+                    FIXED_ONE_BYTE_STRING(env->isolate(), "ECDH")));
+          USE(info->Set(context, env->name_string(),
+                    OneByteString(env->isolate(), curve_name)));
+          USE(info->Set(context, env->size_string(),
+                    Integer::New(env->isolate(), EVP_PKEY_bits(key.get()))));
```
addaleax: So … this mixes repeated use of `USE()` here (which isn’t great because it might mean calling into JS while an exception is pending) with `.FromJust()` above (which also isn’t great because we can do better than crash on exceptions).

I’d either pick `.FromJust()` and stick to that, or do proper error handling and go with
`if (info->Set(...).IsNothing()) return MaybeLocal<Value>();`.

src/node_quic_session.cc
```diff
@@ -165,6 +165,540 @@ void QuicSessionConfig::GeneratePreferredAddressToken(ngtcp2_cid* pscid) {
...
+    QuicSession* session,
+    SecureContext* ctx,
+    ngtcp2_crypto_side side,
+    uint32_t options) :
+    session_(session),
+    side_(side),
+    in_tls_callback_(false),
+    in_key_update_(false),
+    in_ocsp_request_(false),
+    in_client_hello_(false),
```
addaleax: Can you use default initializers in the class definition itself instead for all constants?

src/node_quic_session.cc
```diff
@@ -1143,6 +1574,111 @@ int IsEmpty(const ngtcp2_vec* vec, size_t cnt) {
...
+      if (req.addrinfo == nullptr)
+        return false;
+
+      dest->addrlen = req.addrinfo->ai_addrlen;
+      memcpy(dest->addr, req.addrinfo->ai_addr, req.addrinfo->ai_addrlen);
+      uv_freeaddrinfo(req.addrinfo);
+      break;
+    }
+    case QUIC_PREFERRED_ADDRESS_IGNORE:
+      // Fall-through
```
addaleax: This does not fall through :)

```suggestion
```

src/node_quic_session.cc
```diff
@@ -1535,36 +2213,41 @@ void QuicSession::UpdateIdleTimer() {
...
+  for (size_t n = 0; n < nsv; n++) {
+    USE(versions->Set(
+        env()->context(), n,
+        Integer::New(env()->isolate(), sv[n])));
+  }
+
+  Local<Array> supportedVersions = Array::New(env()->isolate(), 1);
+  USE(supportedVersions->Set(
+      env()->context(), 0,
+      Integer::New(env()->isolate(), NGTCP2_PROTO_VER)));
```
addaleax: style nit: `snake_case` for local variables

Substantial nit: a) repeated use of `USE()` is not great, as mentioned above, and b) we basically always prefer the 3-argument variant of `Array::New()` that takes an array of `Local<Value>`s for creating arrays this way.

test/pummel/test-heapdump-quic.js
```diff
@@ -69,62 +69,76 @@ server.on('session', common.mustCall((session) => {
...
           { node_name: 'QuicServerSession', edge_name: 'wrapped' },
-          { node_name: 'Node / rx_secret', edge_name: 'rx_secret' },
-          { node_name: 'Node / tx_secret', edge_name: 'tx_secret' },
+          { node_name: 'Node / QuicCryptoContext', edge_name: 'crypto_context'},
+          // { node_name: 'Node / rx_secret', edge_name: 'rx_secret' },
+          // { node_name: 'Node / tx_secret', edge_name: 'tx_secret' },
+          // { node_name: 'Node / QuicBuffer', edge_name: 'initial_crypto' },
+          // { node_name: 'Node / QuicBuffer',
+          //   edge_name: 'handshake_crypto' },
+          // { node_name: 'Node / QuicBuffer', edge_name: 'app_crypto' },
```
addaleax: These can be removed, right?

src/node_quic_crypto.cc
jasnell: I'll go with FromJust for now but we can revisit this later

jasnell:
Ping @addaleax

test/pummel/test-heapdump-quic.js
```diff
@@ -69,62 +69,76 @@ server.on('session', common.mustCall((session) => {
...
     ], { loose: true });
 
-    state.validateSnapshotNodes('Node / QuicServerSession', [
+    state.validateSnapshotNodes('Node / QuicSession', [
       {
         children: [
           { node_name: 'QuicServerSession', edge_name: 'wrapped' },
-          { node_name: 'Node / rx_secret', edge_name: 'rx_secret' },
-          { node_name: 'Node / tx_secret', edge_name: 'tx_secret' },
+          { node_name: 'Node / QuicCryptoContext', edge_name: 'crypto_context'},
```
danbev: Nit: A space if needed before the closing bracket:
```console
76:79  error  A space is required before '}'  object-curly-spacing
```

test/pummel/test-heapdump-quic.js
```diff
@@ -69,62 +69,76 @@ server.on('session', common.mustCall((session) => {
...
-    ], { loose: true });
-
-    state.validateSnapshotNodes('Node / QuicClientSession', [
+      },
       {
         children: [
           { node_name: 'QuicClientSession', edge_name: 'wrapped' },
-          { node_name: 'Node / rx_secret', edge_name: 'rx_secret' },
-          { node_name: 'Node / tx_secret', edge_name: 'tx_secret' },
+          { node_name: 'Node / QuicCryptoContext', edge_name: 'crypto_context'},
```
danbev: Nit: space before the closing bracket.

test/pummel/test-heapdump-quic.js
jasnell:
```suggestion
          { node_name: 'Node / QuicCryptoContext', edge_name: 'crypto_context' },
```

test/pummel/test-heapdump-quic.js
jasnell:
```suggestion
          { node_name: 'Node / QuicCryptoContext', edge_name: 'crypto_context' },
```

jasnell:
Landed!

PR #169, opened 2019-10-11
test: enable test_quic_verifyhostnameidentity.cc

danbev:
This commit adds a suggestion for enabling the cctest
test_quic_verifyhostnameidentity.cc.

##### Checklist


test/cctest/test_quic_verifyhostnameidentity.cc
```diff
@@ -1,6 +1,6 @@
 
 #include "base_object-inl.h"
-#include "node_quic_crypto.h"
+#include "../src/node_quic_crypto.cc"  // NOLINT(build/include)
```
addaleax: Could we maybe add the necessary overload back to `node_quic_crypto.h`? If I had noticed that it’s still there I would have done that

danbev: @jasnell Do you have an input on this, would it be alright to move this function back again like @addaleax suggested?

jasnell: Yeah that's fine!

addaleax: Just to avoid confusion – I wasn’t suggesting to move the function definition, but rather to add a declaration for it to the header. That should be enough, I think.

addaleax:
Landed in 46730e899d1f5ad07700e92979b78216a123c6a4 :)

Issue #174, opened 2019-10-16
How to start running Node with QUIC protocol locally


OmerHerera:
I don't want this issue will look a complain, you are doing a great job in the project, keep the great work 🎊 

**Is your feature request related to a problem? Please describe.**

For new contributors/potential like me, it's very difficult to understand which steps I need to run to get nodejs running my machine.
How I can run the tests?
How I can see QUICK protocol running?
I want to compare HTTP/2 against HTTP/3(QUIC), so I need a nodejs server that supports it.  

**Describe the solution you'd like**

I will like to have documentation with the steps required 

**Describe alternatives you've considered**

Include a section in README for new/potential contributors

OmerHerera:
Thanks for your quick answer.
I'm trying to build node locally but it's failing, seems that I don't have all the pre-requisites dependencies or the proper versions on my local machine.
Do you know if exists a Docker with all the pre-requisites already installed and I can only mount the code and compile, I'm installing dependencies in my local machine but it seems that still missing or wrong versions.
This will make easier for new contributors like me, that you only pull the image and start building.
Thanks in advance

juanarbol:
I learned how to build Node by reading this file [BUILDING.md](https://github.com/nodejs/node/blob/master/BUILDING.md)

PR #83, opened 2019-08-21
quic: add experimental-quic build flag

jasnell:
Fixes: https://github.com/nodejs/quic/issues/55

##### Checklist

- [ ] `make -j4 test` (UNIX), or `vcbuild test` (Windows) passes
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

PR #175, opened 2019-10-16
doc: add `--experimental-quic` in the doc

trivikr:
Refs: https://github.com/nodejs/quic/issues/174
Refs: https://github.com/nodejs/quic/pull/83

##### Checklist


jasnell:
Currently the build flag does not gate the modifications to openssl but I actually want them to. The reason I went with a build flag for now is so we can land in master soon without worrying too much about the still in flux changes to the openssl patches and the addition of the two new dependency libraries. There's also the long standing and still unresolved question about whether or not it is ok to introduce new unscoped top level modules (e.g. `require('quic')`) and the fact that a module called `quic` already exists in npm. Because of all that, I figured a build flag was generally safer for now. The goal would be to remove the flag as soon as possible and not even require a runtime flag, just have `require('quic')` emit the experimental warning while it is still experimental.

jasnell:
Rather than make this change, could we create a pinned issue that includes the build instructions?

trivikr:
> I assume the idea is that this will not be merged into the main repo when the PR is opened?

Yup, the idea is to revert this change before merging with Node.js core

> Rather than make this change, could we create a pinned issue that includes the build instructions?

+1 for creating a pinned issue

trivikr:
A pinned issue was created at https://github.com/nodejs/quic/issues/183

PR #176, opened 2019-10-16
quic: allow multiple independent .connect() calls

addaleax:
Explicitly allow calling `.connect()` multiple times to generate
separate client sessions on the same socket. This requires making
`kMaybeBind` call its callback even if the socket is already bound.

##### Checklist


lib/internal/quic/core.js
```diff
@@ -727,15 +727,18 @@ class QuicSocket extends EventEmitter {
...
+      if (this.bound)
+        return process.nextTick(callback);
+      else
+        return this.on('ready', callback);
+    }
     // This socket will be in a pending state until it is bound. Once bound,
     // the this[kReady]() method will be called, switching the state to
-    // kSocketBound and notifying the associated sessions
-    // TODO(@jasnell): If the socket is already bound, the callback should
-    // be invoked with an error.
```
addaleax: @jasnell I assume this TODO didn’t exist without a reason … do you remember more about this?

jasnell: ummmm... I think this is no longer relevant, particularly with the changes you've been making here.

addaleax:
Landed in b2197b548a62fd89c66669daa0f430cd42a48a18

PR #177, opened 2019-10-16
quic: continuing work on the http/3 implementation

jasnell:
Continuing work on the http/3 implementation


jasnell:
@addaleax @nodejs/quic ... Making progress on the http3 implementation. This is probably about 90% there. We have headers and data flowing but there's still some session lifecycle stuff to work through. I'll be continuing work on it on Monday

src/node_http_common.h
```diff
@@ -261,7 +261,7 @@ class NgRcBufPointer : public MemoryRetainer {
  public:
   typedef typename T::rcbuf_t rcbuf_t;
   typedef typename T::vector_t vector_t;
-  NgRcBufPointer() {}
+  NgRcBufPointer() : buf_(nullptr) {}
```
addaleax: Can this be `buf_ = nullptr;` in the field definition instead? It’s cleaner and avoid theses changes here

src/node_http_common.h
```diff
@@ -307,7 +307,12 @@ class NgRcBufPointer : public MemoryRetainer {
     return *new (this) NgRcBufPointer(std::move(other));
   }
 
-  ~NgRcBufPointer() { T::dec(get()); }
+  ~NgRcBufPointer() {
+    rcbuf_t* ptr = get();
+    if (ptr) {
```
addaleax:
```suggestion
    if (ptr != nullptr) {
```

src/node_quic_http3_application.cc
```diff
@@ -393,14 +455,28 @@ ssize_t Http3Application::H3ReadData(
...
-    stream->DrainInto(&vec, &count, std::min(veccnt, MAX_VECTOR_COUNT));
-    CHECK_LE(count, MAX_VECTOR_COUNT);
-    if (!stream->IsWritable())
-      *pflags |= NGHTTP3_DATA_FLAG_EOF;
+printf(">>>>>>>>>>>>>\n");
+  stream->DrainInto(&vec, &count, std::min(veccnt, MAX_VECTOR_COUNT));
+  CHECK_LE(count, MAX_VECTOR_COUNT);
+  size_t numbytes = nghttp3_vec_len(vec, count);
+  stream->Commit(numbytes);
+printf("<<<<<<<<<<<<<\n");
```
addaleax: Just commenting to make sure these don’t get lost :)

src/node_quic_http3_application.cc
```diff
@@ -464,9 +540,9 @@ void Http3Application::H3BeginHeaders(
   int64_t stream_id,
   QuicStreamHeadersKind kind) {
   QuicStream* stream = FindOrCreateStream(stream_id);
-  if (!stream)
-    return;
-  stream->BeginHeaders(kind);
+  Debug(Session(), "Starting header block for stream %" PRId64, stream_id);
+  if (stream)
```
addaleax:
```suggestion
  if (stream != nullptr)
```

src/node_quic_session.cc
```diff
@@ -611,17 +612,34 @@ bool QuicCryptoContext::UpdateKey(bool initiate) {
...
   IncrementStat(
       1, &session_->session_stats_,
       &QuicSession::session_stats::keyupdate_count);
 
-  if (!UpdateAndInstallKey(session_, &rx_secret_, &tx_secret_))
-    return false;
+  int err = ngtcp2_conn_initiate_key_update(
+             session_->Connection(),
+             uv_hrtime());
+printf("Initiated key update? %d\n", err);
```
addaleax: ditto

test/parallel/test-quic-client-server.js
```diff
@@ -143,7 +143,7 @@ server.on('session', common.mustCall((session) => {
     // Should not error and should return true... also shouldn't
     // cause anything else to fail.
 
-    assert(session.updateKey());
+    setTimeout(() => assert(session.updateKey()), 500);
```
addaleax: Why the timeout? It sounds like something that could make this test flaky?

jasnell: this is temporary for now. With the update to ngtcp2, we can't perform the key update at this point in the lifecycle because we're still inside the ngtcp2 callback at this point. I'll be removing the key update from this test and moving it into it's own test soonish

jasnell:
@addaleax @nodejs/quic ... this is ready for review! should be ready to land.

src/env.h
```diff
@@ -322,7 +322,7 @@ constexpr size_t kFsStatsBufferLength =
   V(promise_string, "promise")                                                 \
   V(pubkey_string, "pubkey")                                                   \
   V(query_string, "query")                                                     \
-  V(quic_alpn_string, "h3-22")                                                 \
+  V(quic_alpn_string, "h3-24")                                                 \
```
addaleax: While working on #200 I ran into 

```
=== release test-quic-binding ===                   
Path: parallel/test-quic-binding

+ 'ff000018'
- 'ff000017'
```

Is that something that needs to be updated in this PR too?

src/node_http_common.h
```diff
@@ -270,16 +270,16 @@ class NgRcBufPointer : public MemoryRetainer {
   SET_MEMORY_INFO_NAME(NgRcBufPointer)
   SET_SELF_SIZE(NgRcBufPointer)
 
-  explicit NgRcBufPointer(rcbuf_t* buf) {
+  explicit NgRcBufPointer(rcbuf_t* buf) : buf_(nullptr) {
```
addaleax: I think all of the `:  buf_(nullptr)` additions can be removed now

src/node_quic_http3_application.cc
```diff
@@ -477,11 +557,12 @@ bool Http3Application::H3ReceiveHeader(
     uint8_t flags) {
   // Protect against zero-length headers
   if (!IsZeroLengthHeader(name, value)) {
+    Debug(Session(), "Receiving header for stream %" PRId64, stream_id);
     QuicStream* stream = Session()->FindStream(stream_id);
     if (stream) {
       if (token == NGHTTP3_QPACK_TOKEN__STATUS) {
         nghttp3_vec vec = nghttp3_rcbuf_get_buf(value);
-        if (memcmp(vec.base, "1", 1))
+        if (memcmp(vec.base, "1", 1) == 0)
```
addaleax:
```suggestion
        if (vec.base[0] == '1')
```

src/node_quic_session.cc
```diff
@@ -2125,12 +2145,13 @@ void QuicSession::StreamReset(
...
-  if (expiry < now || timeout == 0) timeout = 1;
-  Debug(this, "Updating idle timeout to %" PRIu64, timeout);
-  idle_->Update(timeout);
+  // uint64_t now = uv_hrtime();
+  // uint64_t expiry = ngtcp2_conn_get_idle_expiry(Connection());
+  // uint64_t timeout = expiry > now ? (expiry - now) / 1000 : 1;
+  // if (timeout == 0) timeout = 1;
+  // Debug(this, "Updating idle timeout to %" PRIu64, timeout);
+  // idle_->Update(timeout);
+  idle_->Update(10000);  // TODO(@jasnell): Fix
```
addaleax: Did this pass tests? (I think I remember this causing a bit of trouble with those…) Or, more generally, is this something that should be addressed before the PR lands?

test/parallel/test-quic-client-server.js
addaleax: I think I also remember this test failing because the session wasn’t alive at this point anymore?

src/env.h
jasnell: Yes. Will do.

src/node_quic_session.cc
jasnell: oh, right right. Yes, thank you. With the update to ngtcp2 there's something wonky about the calculation so I had to temporarily disable. Yes, this needs to be fixed before this lands.

jasnell: Interesting... the timeout calculation is working just fine on Linux but not on Windows... hmmmm

jasnell: Ok, confirmed... yeah, this is very weird... On linux, the expiry calculation works as expected, with the expiry time reported by the ngtcp2 API returning a timestamp greater than the current time returned by uv_hrtime()... on windows, however, the expiry time is always less than uv_hrtime() so the timeout is setting as 1..

jasnell: It's happening with DefaultApplication and Http3Application. It's something that has changed directly within ngtcp2.

jasnell: Figured it out. The idle_timeout was being set assuming milliseconds but the code was expecting nanoseconds. It was throwing the calculation off.

jasnell:
Ok, the idle timeout and updateKey bits have been addressed (updateKey by removing it from the test for now)

addaleax:
Current test failures that I’m seeing locally:

```
=== release test-quic-client-server ===                                       
Path: parallel/test-quic-client-server
(node:12422) internal/test/binding: These APIs are for internal testing only. Do not use them.
(node:12422) ExperimentalWarning: QUIC protocol support is experimental and not yet supported for production use
assert.js:90
  throw new AssertionError(obj);
  ^

AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 2

    at QuicServerSession.<anonymous> (/home/sqrt/src/node/quic/test/parallel/test-quic-client-server.js:216:12)
    at QuicServerSession.<anonymous> (/home/sqrt/src/node/quic/test/common/index.js:372:15)
    at QuicServerSession.emit (events.js:215:7)
    at processTicksAndRejections (internal/process/task_queues.js:75:11) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 0,
  expected: 2,
  operator: 'strictEqual'
}
Command: out/Release/node --expose-internals /home/sqrt/src/node/quic/test/parallel/test-quic-client-server.js
=== release test-quic-idle-timeout ===                                   
Path: parallel/test-quic-idle-timeout
(node:12455) ExperimentalWarning: QUIC protocol support is experimental and not yet supported for production use
Command: out/Release/node /home/sqrt/src/node/quic/test/parallel/test-quic-idle-timeout.js
--- TIMEOUT ---
=== release test-quic-ipv6only ===                        
Path: parallel/test-quic-ipv6only
(node:12462) ExperimentalWarning: QUIC protocol support is experimental and not yet supported for production use
Command: out/Release/node /home/sqrt/src/node/quic/test/parallel/test-quic-ipv6only.js
--- TIMEOUT ---
[00:20|% 100|+  24|-   3]: Done   
```

Is that something worth looking into before merging? If anything doesn’t reproduce for you I can try to do so myself.

jasnell:
The test is working fine for me on Windows, will try on Linux

jasnell:
Working for me on Linux also

addaleax:
Okay, then don’t let that hold you up and I’ll try to debug the failures in the meantime

jasnell:
Landed

PR #178, opened 2019-10-16
quic: implement memory tracking for QuicApplication

addaleax:
This addresses a TODO comment.

##### Checklist

- [x] `make -j4 test` (UNIX), or `vcbuild test` (Windows) passes
- [x] tests and/or benchmarks are included
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

PR #179, opened 2019-10-16
quic: fix linting and tests

addaleax:
Once again, linter and tests have been failing because of recent
changes. :)

##### Checklist

### DIFF
--- a/src/node_quic_http3_application.cc
+++ b/src/node_quic_http3_application.cc
@@ -1,11 +1,11 @@
 #include "node.h"
 #include "debug_utils.h"
-#include "node_http_common.h"
+#include "node_mem-inl.h"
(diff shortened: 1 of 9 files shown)


src/node_quic_http3_application.cc
```diff
@@ -1,11 +1,11 @@
 #include "node.h"
 #include "debug_utils.h"
-#include "node_http_common.h"
+#include "node_mem-inl.h"
```
addaleax: (This is to match #178)

jasnell:
ah... I keep forgetting to lint the c bits on linux instead of windows. `vcbuild lint` misses these bits

jasnell:
landed!

PR #180, opened 2019-10-16
quic: rename shutdown to reset for RESET_STREAM methods

addaleax:
`Shutdown()` is already used in our own `StreamBase` API, and
in particular `QuicStream::Shutdown()` shadowed
`StreamBase::Shutdown()`; the former cleanly shuts down the
writable side of a stream, whereas the latter indicates an
abrupt termination through a `RESET_STREAM` frame.

##### Checklist


jasnell:
Landed

PR #186, opened 2019-10-26
quic: make `maxConnectionsPerHost` works

oyyd:
It seems that `Hash` class for `unorderred_map` is used to decide which
bucket to put the key. But keys that are different pointers with the
same hash will still be regarded as different keys so that we still need
to specify a `std::equal_to` to compare the key correctly.

After this PR however, when `maxConnectionsPerHost` is exceeded, the new session will
still be emitted on the server side and won't be destroied automatically.
We need to figure out the reason and fix it.

##### Checklist

### DIFF
--- a/src/node_quic_socket.h
+++ b/src/node_quic_socket.h
@@ -250,8 +250,8 @@ class QuicSocket : public AsyncWrap,
   // value reaches the value of max_connections_per_host_,
   // attempts to create new connections will be ignored
   // until the value falls back below the limit.
-  std::unordered_map<const sockaddr*, size_t, SocketAddress::Hash>
-      addr_counts_;
+  std::unordered_map<const sockaddr*, size_t, SocketAddress::Hash,
+      SocketAddress::Compare> addr_counts_;
 
   // The validated_addrs_ vector is used as an LRU cache for
   // validated addresses only when the VALIDATE_ADDRESS_LRU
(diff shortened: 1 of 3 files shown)


addaleax:
@oyyd Thanks, I knew something was bugging me about this code :sweat_smile: 

This approach works only as well as the hash function, though; if two hosts happen to end up with the same hash, then they will be counted as one result.

I’m not sure, but apart from accidentally limiting hosts through this, this might enable an attack scenario where an attacker specifically picks its sending address so that it matches the hash of another host and thus forbids that other, non-malicious host from opening connections. (Picking an address should be easy enough at least with IPv6 hosts that are given address ranges rather than single addresses.)

If we want to follow the original intention of the code, which is to perform lookups based on the full address, then what we probably want to do is to provide an equality comparison in addition to the hash function, as a 4th argument to `unordered_map`, that compares socket addresses properly according to their contents, rather than the pointers.

oyyd:
That's ture.. I think I have misunderstood the original intention (and have been not familiar enough with `unorderred_map`). Will push another commit later.

jasnell:
Yep, this was done initially as a dirty hack :-) ... it definitely has needed a revisit.

The original intent is definitely to provide lookup on the full address. The irony with this specific implementation is that this needs to be in place to prevent an attack vector but the hacky current implementation *is* an attack vector ;-)

addaleax:
Landed in 32954ff91316829ebaba5d756e694a0b341f507d, nice PR!

Issue #181, opened 2019-10-16
openStream() is sync but can’t be


addaleax:
Calling `openStream()` immediately after `client.connect()` crashes because the underlying session handle isn’t set yet, but it requires the handle for synchronous operation.

I’m not sure how to best approach this from an API perspective. There isn’t really any way we can fully make the current API work, I think, because at least the stream IDs are something we don’t really know before the underlying `QuicSession` is set up.

Maybe this should just throw an exception in that case?

jasnell:
So this one is going to be a bit complicated. It should be absolutely possible to call `openStream()` immediately after `client.connect()` with the `QuicStream` remaining in a pending state until the underlying `QuicSession` handle has been set. While in the pending state, it should be possible to set headers (when relevant) and even begin queuing data to send (with the Duplex operating in corked mode). If the `QuicSession` is a client, then any pending `QuicStream` instances should actually become 0RTT streams, with initial request headers being transmitted as part of the 0RTT initial handshake. None of this piece is implemented yet tho because I just wasn't sure how we were going to do so. A `QuicStream` would remain in the pending state until the `QuicSession` is established and is in a state where it can uncork the Duplex and begin flowing data.

Note, ngtcp2 will allow streams to be opened during the handshake, and will assign id's to those, with the assumption that those are being used as 0RTT streams.

jasnell:
The flow on this then... 

* Any time after the creation of a JavaScript `QuicClientSession` or `QuicServerSession`, a JavaScript `QuicStream` may be created.

* If, at the time the `QuicStream` is created, the parent `QuicSession` does not have a handle set, the `QuicStream` exists in a `Pending` state, state changing operations in JavaScript are queued. Note that this should never actually happen with `QuicServerSession` instances which always have the handle created first, so the rest assumes `QuicClientSession` only...

* If, at the time the `QuicClientSession` has it's handle set, the TLS handshake is still initiating, any existing pending `QuicStream` instances would be created and moved immediately to a `ORTT` state, which allows some initial data to be transmitted within very strict limits set by the QUIC spec.

* Once the TLS handshake completes, any existing `QuicStream` instances would move to a `Ready` state.

This model allows the `QuicStream` to be created synchronously but requires a bit more complicated state tracking at the native layer than what is currently implemented.

addaleax:
[...]

PR #187, opened 2019-10-26
quic: add pending state for streams

addaleax:
This resolves the situation in which `openStream()` is called before
the underlying native `QuicSession` handle exists.

Fixes: https://github.com/nodejs/quic/issues/181

##### Checklist


test/parallel/test-quic-openstream-pending.js
```diff
@@ -0,0 +1,70 @@
...
+  const req = client.connect({
+    address: 'localhost',
+    port: server.address.port
+  });
+
+  const stream = req.openStream({ halfOpen: true });
+  stream.end('Hello!');
+
+  assert.strictEqual(stream.pending, true);
+  stream.on('ready', common.mustCall(() => {
```
jasnell: In the `ready` event, I'm wondering if the stream ID shouldn't be passed off to the callback, just as a helpful bit of context.

addaleax: I’m not sure if that really provides more context – it is available as `stream.id` (or `this.id`) in the event handler – but I have added a note about it in the documentation.

trivikr: Resolved in https://github.com/nodejs/quic/pull/187/commits/340eaceaebf374dcdd2cf41f920974c30cdb85d1

addaleax:
Landed in 0e863697940c41e92750220c78049b67c821e68c

PR #191, opened 2019-11-06
quic: remove unused functional include in session

danbev:
This commit removes the include of the functional header from
node_quic_session.h and node_quic_session.cc as the are not currently
used.

##### Checklist


danbev:
Landed in 70ffbd0b3680dbadaeb22a8e2673b1cd25e97419.

PR #193, opened 2019-11-08
quic: remove algorithm include

danbev:
This commit removes the include of the algorithm header as I cannot
find any usage of it.

##### Checklist


danbev:
> What is `vector` used for here?

Sorry, I misread the code and it does not use vector, but [map and iterator](https://github.com/nodejs/quic/blob/f0b2d204e21a1552409e2f8a3cc3451a6bdb39e9/src/node_quic_session-inl.h#L60). 

Unless I'm mistaken `std::find` is from [map](https://en.cppreference.com/w/cpp/container/map) and the free function `std::end` from [iterator](https://en.cppreference.com/w/cpp/iterator/end), and iterator is included via map. 
 
In this case we know that `streams_` is of type map and perhaps we could use the member function `end` instead:
```c++
return streams_.find(id) != streams_.end();
```
My understanding of the motivation for using the free function `std::end` is that it would be applicable on containers that don't provide an `end` function.  But in this case I'm not sure that applies. Any thought on this?

danbev:
Landed in cb8c0b7000539a7798a761dad2ea9beb813ad317.

PR #194, opened 2019-11-08
test: improve code coverage on quic implementation

juanarbol:
##### Checklist

### DIFF
--- a/test/parallel/test-quic-quicsocket.js
+++ b/test/parallel/test-quic-quicsocket.js
@@ -139,12 +139,14 @@ socket.on('ready', common.mustCall(() => {
 
 socket.on('close', common.mustCall(() => {
   const expectedError = { code: 'ERR_QUICSOCKET_DESTROYED' };
-  assert.throws(() => socket.setTTL(1), expectedError);
-  assert.throws(() => socket.setMulticastTTL(1), expectedError);
-  assert.throws(() => socket.setBroadcast(true), expectedError);
-  assert.throws(() => socket.setMulticastLoopback(), expectedError);
-  assert.throws(() => socket.setMulticastInterface(true), expectedError);
-  assert.throws(() => socket.addMembership('foo', 'bar'), expectedError);
-  assert.throws(() => socket.dropMembership('foo', 'bar'), expectedError);
+  const message = (method) => `Cannot call ${method} after a QuicSocket has been destroyed`
+
+  assert.throws(() => socket.setTTL(1), { ...expectedError, message: message('setTTL') });
+  assert.throws(() => socket.setMulticastTTL(1), { ...expectedError, message: message('setMulticastTTL') });
+  assert.throws(() => socket.setBroadcast(true), { ...expectedError, message: message('setBroadcast') });
+  assert.throws(() => socket.setMulticastLoopback(), { ...expectedError, message: message('setMulticastLoopback') });
+  assert.throws(() => socket.setMulticastInterface(true), { ...expectedError, message: message('setMulticastInterface') });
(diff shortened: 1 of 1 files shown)


juanarbol:
I've a question, I think this tests could be more precise, for example in [core.js](https://github.com/nodejs/quic/blob/master/lib/internal/quic/core.js#L1213), the threw error is `new ERR_QUICSOCKET_DESTROYED('setBroadcast')`, how could I check not only the code (`ERR_QUICSOCKET_DESTROYED` in this case) and check param sent to error (`ERR_QUICSOCKET_DESTROYED('setBroadcast')`).

addaleax:
> I've a question, I think this tests could be more precise, for example in [core.js](https://github.com/nodejs/quic/blob/master/lib/internal/quic/core.js#L1213), the threw error is `new ERR_QUICSOCKET_DESTROYED('setBroadcast')`, how could I check not only the code (`ERR_QUICSOCKET_DESTROYED` in this case) and check param sent to error (`ERR_QUICSOCKET_DESTROYED('setBroadcast')`).

@juanarbol You could check the `message:` property, similar to the `code:` property?

juanarbol:
Yes, I could compare messages, I've made this, it works, but I wanna know what do you think.

```js
const expectedError = { code: 'ERR_QUICSOCKET_DESTROYED' };
const message = (method) => `Cannot call ${method} after a QuicSocket has been destroyed`
assert.throws(() => socket.setTTL(1), { ...expectedError, message: message('setTTL') });
```

PS: Not all of them has messages.

addaleax:
@juanarbol Yeah, that looks just fine to me :)

juanarbol:
I've found (I guess) a typo on `setServerBusy` method, so I opened #195

juanarbol:
This didn't pass lint, working on it!

test/parallel/test-quic-quicsocket.js
```diff
@@ -136,3 +136,42 @@ socket.on('ready', common.mustCall(() => {
...
 }));
+
+socket.on('close', common.mustCall(() => {
+  const err = { code: 'ERR_QUICSOCKET_DESTROYED' };
+  const message = (method) => `Cannot call ${method} ` +
+                              'after a QuicSocket has been destroyed';
+
+  assert.throws(
+    () => socket.setTTL(1),
+    { ...err, message: message('setTTL') }
```
lundibundi: Nit: perhaps this would be simpler and avoids duplication
```javascript
const makeError = (method) => ({
  code: 'ERR_QUICSOCKET_DESTROYED',
  message: `Cannot call ${method} after a QuicSocket has been destroyed`;
});

// and then
  assert.throws(
    () => socket.setMulticastTTL(1),
    makeError('setMulticastTTL'),
  );

// or probably
assert.throws(() => socket.setMulticastTTL(1), makeError('setMulticastTTL'));
```

juanarbol: Yeap! Give me few hours to follow that suggestion

test/parallel/test-quic-quicsocket.js
```diff
@@ -136,3 +136,43 @@ socket.on('ready', common.mustCall(() => {
...
+    () => socket.addMembership('foo', 'bar'),
+    makeError('addMembership')
+  );
+  assert.throws(
+    () => socket.dropMembership('foo', 'bar'),
+    makeError('dropMembership')
+  );
+  assert.throws(
+    () => socket.setServerBusy(true),
+    makeError('setServerBusy')
```
juanarbol: @lundibundi Thanks for your suggestion! It's more elegant, I've splitted throws into two lines because of the linter.

danbev:
Landed in 0ffa2ea729afb533504f58618f4a7a2ee67f24f4.

Issue #161, opened 2019-10-09
TODO: Disable QUIC when using shared_openssl


jasnell:
The QUIC implementation requires a modified version of that statically compiled openssl. QUIC should not be available when using shared_openssl. Need to make sure it is disabled in the build when configured to build with shared_openssl

PR #196, opened 2019-11-12
build: fail when quic is enabled with shared_openssl in configure

lundibundi:
Fixes: https://github.com/nodejs/quic/issues/161

/cc @jasnell 

##### Checklist

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1120,6 +1120,9 @@ def configure_node(o):
     o['variables']['debug_nghttp2'] = 'false'
 
   if options.experimental_quic:
+    if options.shared_openssl:
+      raise Exception('QUIC requires modified version of OpenSSL and cannot be'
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1120,6 +1120,9 @@ def configure_node(o):
     o['variables']['debug_nghttp2'] = 'false'
 
   if options.experimental_quic:
+    if options.shared_openssl:
+      raise Exception('QUIC requires modified version of OpenSSL and cannot be'
+                      ' enabled with --shared_openssl.')
```
danbev: Should this perhaps be `--shared-openssl` instead, as that is the name of the option passed on the command line?

danbev:
Landed in ac8834d1a351e6989ad383779e068bf2759824f3.

PR #198, opened 2019-11-19
quic: ensure callbacks of QuicSocket.connect() get called

oyyd:
The callbacks of `QuicSocket.connect()` won't get called after
`QuicSocket` binding. This PR calls `QuicSession[kReady]()` directly
when `QuicSocket`  bound to fix this issue.

This PR also modify `SocketAddress::Hash` and `SocketAddress::Compare`
to accept values instead of pointers because `SocketAddress` might
get freed firstly which would cause that the values aren't safe.
For example, the test added is likely to abort before this PR.

##### Checklist


src/node_quic_socket.h
```diff
@@ -250,7 +250,7 @@ class QuicSocket : public AsyncWrap,
   // value reaches the value of max_connections_per_host_,
   // attempts to create new connections will be ignored
   // until the value falls back below the limit.
-  std::unordered_map<const sockaddr*, size_t, SocketAddress::Hash,
+  std::unordered_map<const sockaddr, size_t, SocketAddress::Hash,
```
addaleax: Unfortunately, this doesn’t quite work, because now we store a key of type `sockaddr`; but `sockaddr_in` and `sockaddr_in6` are “subclasses” of `sockaddr`, and so a key that is just `sockaddr` won’t be able to actually hold the full address.

`sockaddr_storage` would probably the best key type, although that comes with a bit of overhead as it’s supposed to also be able to hold UNIX socket paths which are much longer than IPv4/IPv6 addresses… I’d still use it, and I’m not sure whether QUIC over UNIX sockets is technically impossible or something we should restrict?

oyyd: Thanks for the comment! I would like to give `sockaddr_storage` a try.

src/node_quic_socket.cc
```diff
@@ -582,9 +582,13 @@ BaseObjectPtr<QuicSession> QuicSocket::AcceptInitialPacket(
   // Check to see if the number of connections for this peer has been exceeded.
   // If the count has been exceeded, shutdown the connection immediately
   // after the initial keys are installed.
-  if (GetCurrentSocketAddressCounter(addr) >= max_connections_per_host_) {
-    Debug(this, "Connection count for address exceeded");
-    initial_connection_close = NGTCP2_SERVER_BUSY;
+  {
+    sockaddr_storage storage;
+    memcpy(&storage, addr, SocketAddress::GetAddressLen(addr));
+    if (GetCurrentSocketAddressCounter(&storage) >= max_connections_per_host_) {
```
oyyd: Try to use `sockaddr_storage` when possible but still have to make a copy here.

oyyd:
@addaleax PTAL

src/node_quic_socket.h
```diff
@@ -172,9 +172,9 @@ class QuicSocket : public AsyncWrap,
       QuicCID* scid,
       const sockaddr* addr);
 
-  void IncrementSocketAddressCounter(const sockaddr* addr);
-  void DecrementSocketAddressCounter(const sockaddr* addr);
-  size_t GetCurrentSocketAddressCounter(const sockaddr* addr);
+  void IncrementSocketAddressCounter(const sockaddr_storage* addr);
+  void DecrementSocketAddressCounter(const sockaddr_storage* addr);
+  size_t GetCurrentSocketAddressCounter(const sockaddr_storage* addr);
```
addaleax: I think all the APIs should keep taking `const sockaddr*` or `const sockaddr&`, because they should still work when a `sockaddr_in*` or `sockaddr_in6` is passed in (which are not as large as `sockaddr_storage`)

src/node_quic_socket.cc
addaleax: Do you know why? That seems like it might hide a bug…

oyyd: Because the `addr` is a `sockaddr*` from uv and I think it's not safe to use its value as an `sockaddr_storage` here.

src/node_quic_socket.h
oyyd: I took `const sockaddr_storage*` here to avoid creating `sockaddr_storage` from `sockaddr` as we can't `reinterpret_cast` them directly. Not sure if there are better solutions.

oyyd:
Also cc @jasnell @danbev for more reviews.

jasnell:
Heads up, this will need to be rebased

Issue #185, opened 2019-10-20
Add qlog support


rmarx:
**Is your feature request related to a problem? Please describe.**
It is generally difficult to debug or gain insight into QUIC and HTTP/3 purely from text-based, unstructured logs. Complex issues with long running connections or during the particulars of connection setup are hard to track down and reason about. More insight aids developers first, later students and researchers. 

**Describe the solution you'd like**
For some time, we have been working on "qlog", a standardized structured logging format for QUIC (see https://github.com/quiclog/internet-drafts). It is basically a well-defined JSON file with specified event names and associated data. This format is now supported by over half of the current QUIC implementations and was just today added to ngtcp2 (https://github.com/ngtcp2/ngtcp2/commit/22d9c6d278aa02b49da01305190114834175c9ef). Outputting this standard format allows you to easily use tools like "qvis" (https://qvis.edm.uhasselt.be) for visualization. More tools will follow in the future. 

It would be great if NodeJS's QUIC integration also supports qlog. Initially, it could be just to allow setting the necessary ngtcp2 parameters (see commit above) so qlog files with internal state are written. Eventually, I would hope NodeJS could add additional events from the top-level application into the qlog files as well.

jasnell:
This has definitely been on my todo list to dig into. Once we're a bit further along we'll have to explore how best to support this. It will most likely be similar to how we currently support TLS keylogging... that is, we will generate events on the javascript side with the necessary information and it will be the responsibility of user code to push that out to files. That will give the user the tools they need to enable it but will keep them in complete control over the actual logging.

PR #200, opened 2019-11-22
quic: add qlog support

addaleax:
Fixes: #185

This is as WIP as #177 I assume, as this also requires updating ngtcp2 (and the first commits here are taken from there).

##### Checklist


jasnell:
nice! #177 should be ready to land by Wednesday.

doc/api/quic.md
```diff
@@ -71,6 +71,8 @@ added: REPLACEME
   * `maxConnectionsPerHost` {number} The maximum number of inbound connections
     per remote host. Default: `100`.
   * `port` {number} The local port to bind to.
+  * `qlog` {boolean} Whether to emit ['qlog'][] events for incoming sessions.
+    (For outgoing client sessions, set `client.qlog`.) Default: `false`.
```
jasnell: Does this need to be here or can we base it solely on the presence of the `qlog` event handler the way we do with things like the keylog support?

addaleax:
> nice! #177 should be ready to land by Wednesday (at the latest)

Yeah, there is one more issue here that I think I’ve narrowed down quite a bit … namely, exceptions from nested `MakeCallback`s don’t seem to trigger the fatal exception handler, so exceptions from a `'qlog'` handler currently go unnoticed and the process just continues (and crashes due to some async hooks expectations around uncaught exceptions).

It’s probably not directly related to this PR, but it should be fixed before merging this; I’ve spent some time digging into it yesterday, but it’s not really a trivial issue…

doc/api/quic.md
addaleax: @jasnell My understanding is that we need to pass the native qlog callback to nghttp2 when setting up the connection, i.e. when the native handle object is created, and we can’t always know if there’s `'qlog'` event handlers at that point?

jasnell: Unless there is a significant performance hit associated with adding the callback I would always add it and include the logic in the qlog handler to check to see if a listener has been added.

addaleax: @jasnell I’ll try to measure the perf impact, but my personal expectation would be that it’s not trivial.

The bigger issue, though, is that it’s not clear at what point we’d know whether there’s going to be a JS event listener or not, and that we’d have to buffer data up until then – which could be indefinitely, I guess?

jasnell:
Landed #177. This will need a rebase :-)

doc/api/quic.md
jasnell: Let's land this as is then but once it lands I'll play around with it a bit more to see if we can reasonably do without the config option.

addaleax:
Yup, rebased :+1:

We can merge this if we want, https://github.com/nodejs/quic/pull/200#issuecomment-558637335 still needs to be addressed but it’s probably going to be mostly unrelated to the contents of this PR anyway

jasnell:
Landed

rmarx:
Oh this is fantastic news! Thank you for the great work!

PR #202, opened 2019-11-26
updateKey() should throws if not TLS Handshake

juanarbol:
##### Checklist

**Note:** this test is unstable, sometimes it's passes, sometimes doesn't, seems that handshake is completed before excepcion is handled, I've tried `setInmediate` and `process.nextTick`, did not work; so I'd love to receive some suggestions or opinions.


jasnell:
The updateKey part of this is going to be moved to it's own separate test very soon

juanarbol:
I get it, how could I help with this?

jasnell:
If you'd like to try putting together that separate test, that would be excellent :-)

juanarbol:
Ok, I will create a new file, named as `test-quic-updatekey,js`, is there something else more than this exception to be tested, to be written on this file?

jasnell:
Heads up, this will need to be rebased

test/parallel/test-quic-updatekey.js
```diff
@@ -0,0 +1,70 @@
...
+  cert,
+  ca,
+  requestCert: true,
+  rejectUnauthorized: false,
+  alpn: kALPN,
+  maxCryptoBuffer: 4096,
+});
+server.on('session', common.mustCall((session) => {
+  process.nextTick(() => {
+    assert.throws(() => session.updateKey(), {
```
jasnell: So, investigating further... the issue around updateKey is that a key update can only be triggered (a) only after the TLS handshake has been confirmed and (b) while we're not in the middle of another key update. The TLS handshake is confirmed as soon as we receive the first 1RTT bytes from the peer but ngtcp2 does not really give us a notification at that point so we're going to have to guess somewhat. I'll be working through the semantics on that and will be implementing a `'handshakeConfirmed'` type of event that will flip the switch that enables `updateKey()` to work. Expect that within the next week I hope

juanarbol: Sounds good @jasnell, tell me if you need some help.

juanarbol:
Closing

PR #205, opened 2019-11-27
quic: more work on http3 and util bits

jasnell:
Beginning another round of http3 updates and some other bits

### DIFF
--- a/src/node_crypto_common-inl.h
+++ b/src/node_crypto_common-inl.h
@@ -5,6 +5,7 @@
 #include "node_crypto.h"
 #include "node_crypto_common.h"
 #include "node.h"
+#include "node_internals.h"
 #include "node_url.h"
 #include "string_bytes.h"
 #include "v8.h"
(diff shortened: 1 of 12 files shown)


jasnell:
@addaleax ... a bit more was added to this, can you take another look?

src/node_crypto_common-inl.h
```diff
@@ -0,0 +1,423 @@
...
+using v8::Context;
+using v8::Integer;
+using v8::Isolate;
+using v8::Local;
+using v8::Object;
+using v8::Value;
+
+namespace crypto {
+
+inline void LogSecret(
```
addaleax: Ditto as in the other PR, `inline` should typically go into the `.h` header file; that way, the compiler knows that the method should be declared inline and warns when it’s used but not defined. The other way around, it’s just the linker that fails, and not necessarily reliably.

That being said, it probably makes sense for any functions here that’s longer than 2 lines to be defined in a `.cc` file.

src/node_crypto_common-inl.h
```diff
@@ -0,0 +1,423 @@
...
+        if (url.flags() & url::URL_FLAGS_CANNOT_BE_BASE ||
+            url.flags() & url::URL_FLAGS_FAILED) {
+          continue;  // Skip this one
+        }
+        map.emplace("uri", url.host());
+      }
+    }
+  }
+  sk_GENERAL_NAME_pop_free(names, GENERAL_NAME_free);
+  bio.reset();
```
addaleax:
```suggestion
```

(This is already cleared by the destructor)

src/node_crypto_common-inl.h
jasnell: Since most of the functions in the inl.h are longer than 2 line, I just moved the entire lot in a cc

jasnell:
Landed

PR #206, opened 2019-12-01
quic: additional sendretry tweaks + bug fix

jasnell:
Additional send-retry handling + a bug fix from the qlog PR

@addaleax ... there was a minor issue from the qlog PR where the ~QuicSession would assert due to lack of a `HandleScope`.

### DIFF
--- a/src/node_quic_session.cc
+++ b/src/node_quic_session.cc
@@ -921,6 +921,7 @@ QuicSession::~QuicSession() {
 
   // Emit the 'close' event in JS. This needs to happen after destroying the
   // connection, because doing so also releases the last qlog data.
+  HandleScope scope(env()->isolate());
   MakeCallback(env()->quic_on_session_destroyed_function(), 0, nullptr);
 }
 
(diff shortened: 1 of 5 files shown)


addaleax:
> @addaleax ... there was a minor issue from the qlog PR where the ~QuicSession would assert due to lack of a HandleScope.

:+1: Do you happen to have a stack trace available for that? That might help prevent this kind of error in the future.

src/node_quic_session.cc
```diff
@@ -1558,6 +1559,21 @@ bool QuicSession::ReceivePacket(
...
+        // QuicSession must be a server and must perform
+        // address validation by sending a Retry packet
+        // then immediately close the connection.
+        if (err == NGTCP2_ERR_RETRY && IsServer()) {
+          // TODO(@jasnell): Need to verify proper ordering of
+          // scid and rcid here
+          Socket()->SendRetry(
+              GetNegotiatedVersion(),
+              &QuicCID(scid()),
+              &QuicCID(rcid()),
```
addaleax: `&`-ing a local variable is usually a sign that the function should be taking a const or rvalue reference, i.e. `const QuicCID&` rather than `QuicCID*` here … does that make sense here?

jasnell: Yep, absolutely. Making the change throughout...

jasnell:
Landed

PR #211, opened 2019-12-04
quic: add connection id strategy

jasnell:
We currently generate connection ID's randomly. There are,
however, alternative schemes being defined. Later we may
allow user code to select alternative strategies for
generating connection ids. This adds the basic mechanism.


addaleax:
Landed in 096f2dae919a4b34e148a93ce9aac1bab7a8cbcf

Issue #62, opened 2019-08-21
TODO: Implement stateless reset


jasnell:
Stateless reset support is going to be a bit tricky. An endpoint sends a stateless reset when it does not have access to connection state any longer, but that means we need to keep track of CIDs and their correspnding stateless reset tokens in a way that can survive a crash (which is ironic to say the least). Need to figure out how to implement this protocol feature.

jasnell:
Currently, stateless reset tokens are generated randomly and not stored, which makes them useless for the intended purpose. In the future, we'll want to go with an alternative approach that allows stateless generation and subsequent verification of reset tokens. This can work in a number of ways but following the pattern of retry token generation would seem to make sense.

PR #215, opened 2019-12-04
quic: stateless reset generate strategy

jasnell:
This is a start to supporting stateless reset tokens per the QUIC 24 spec. Rather than generating tokens randomly, generate them cryptographically based on the DCID and send them when we cannot identify a connection and the received packet has a short header.

Fixes: https://github.com/nodejs/quic/issues/62


src/node_quic_socket.cc
```diff
@@ -150,7 +151,8 @@ QuicSocket::QuicSocket(
     uint64_t retry_token_expiration,
     size_t max_connections_per_host,
     uint32_t options,
-    QlogMode qlog)
+    QlogMode qlog,
+    uint8_t* sessionResetSecret)
```
addaleax:
```suggestion
    uint8_t* session_reset_secret)
```

;)

src/node_quic_socket.cc
```diff
@@ -987,14 +1050,22 @@ void NewQuicSocket(const FunctionCallbackInfo<Value>& args) {
...
 
-  new QuicSocket(
-      env,
-      args.This(),
-      args[0].As<Object>(),
-      retry_token_expiration,
-      max_connections_per_host,
-      options,
-      args[4]->IsTrue() ? QlogMode::kEnabled : QlogMode::kDisabled);
+  uint8_t* sessionResetSecret = nullptr;
```
addaleax: ditto here about the casing :)

That being said … this is read-only data, if I understand correctly, right? In that case it would be helpful to make that more obvious by marking it `const uint8_t*` everywhere and/or using `ArrayBufferViewContents` to read it. The latter allows us to avoid extra allocations for small input, and since the size seems to be fixed at 16, I think using it makes sense.

jasnell:
Landed

PR #216, opened 2019-12-05
quic: a few general cleanups

jasnell:
Use of const refs where appropriate and more idiomatic operator use 

/cc @addaleax


jasnell:
Landed

PR #223, opened 2019-12-07
quic: additional edits and multiple endpoint start

jasnell:
Additional edits + Beginning support for multiple local UDP endpoints

### DIFF
--- a/src/node_sockaddr.h
+++ b/src/node_sockaddr.h
@@ -73,7 +73,23 @@ class SocketAddress {
       SocketAddress* addr = nullptr);
 
   inline static SocketAddress* FromSockName(
+      const uv_tcp_t* handle,
+      SocketAddress* addr = nullptr);
+
+  inline static SocketAddress* FromSockName(
+      const uv_tcp_t& handle,
+      SocketAddress* addr = nullptr);
+
+  inline static SocketAddress* FromSockName(
       uv_udp_t* handle,
+      SocketAddress* addr = nullptr);
+
+  inline static SocketAddress* FromSockName(
+      const uv_udp_t* handle,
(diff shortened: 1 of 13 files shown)


src/node_sockaddr.h
```diff
@@ -72,18 +72,50 @@ class SocketAddress {
       uv_tcp_t* handle,
       SocketAddress* addr = nullptr);
```
addaleax: The four overloads taking a non-const pointer are no longer necessary, right?

jasnell: Yeah, I'll be pulling those back out. This is still very much a WIP

jasnell:
Will get this PR updated tomorrow

jasnell:
There's still a lot to do to get QuicSocket's ready for multiple udp endpoints but this lays the ground work. Will do the rest in a separate PR

src/udp_wrap.h
```diff
@@ -46,7 +46,8 @@ class UDPListener {
   // Called right after data is received from the socket, and includes
   // information about the source address. If `nread` is negative, an error
   // has occurred, and it represents a libuv error code.
-  virtual void OnRecv(ssize_t nread,
+  virtual void OnRecv(uv_udp_t* handle,
```
addaleax: This doesn’t seem like something we should be doing … currently, `UDPListener` is already associated with one `UDPWrapBase` instance, so this should always be same same value and always retrievable

As far as I can tell, this is currently only being used for the `SocketAddress::FromSockName()` call above? I might be missing something, but that one could probably use `udp_->GetSockName()` instead?

jasnell: Ok, will rework it!

jasnell:
Landed

PR #224, opened 2019-12-10
src: fix lint-cpp issues

danbev:
##### Checklist


src/node_sockaddr-inl.h
```diff
@@ -12,7 +12,7 @@
 
 namespace node {
 
-namespace {
+namespace {  // NOLINT(build/namespaces)
```
addaleax: Can we remove the namespace here instead? The linter complains for a reason :)

danbev: Absolutely, updated now.

addaleax:
Landed in b3b9c0e438654f0cef504c46598ccc10af15c1f8 :)

PR #230, opened 2019-12-16
quic: multi-endpoint and quicpacket

jasnell:
This accomplishes two important things:

1. Significant simplification of sendbuf and txbuf (by removing those). Adds `QuicPacket` to simplify serialization and flow of outbound packet data.

2. Adds `QuicEndpoint` and multi-endpoint support (not yet fully tested)


jasnell:
@addaleax .. this is ready to review.

src/node_quic_default_application.cc
```diff
@@ -143,12 +144,12 @@ bool DefaultApplication::SendStreamData(QuicStream* stream) {
 
   for (;;) {
     Debug(stream, "Starting packet serialization. Remaining? %d", remaining);
-    MallocedBuffer<uint8_t> dest(Session()->GetMaxPacketLength());
+    auto packet = CreateStreamDataPacket();
```
addaleax: It’s easier to read if the type is spelled out ;)

src/node_quic_socket.cc
```diff
@@ -134,21 +134,109 @@ void JSQuicSocketListener::OnServerBusy(bool busy) {
...
 void JSQuicSocketListener::OnDestroy() {
   // Do nothing here.
 }
 
+QuicEndpoint::QuicEndpoint(
+    Environment* env,
+    Local<Object> wrap,
+    QuicSocket* listener,
+    Local<Object> udp_wrap) :
+    BaseObject(env, wrap),
```
addaleax: mini style nit: `:` goes on the line of the first initializer

src/node_quic_socket.h
```diff
@@ -88,12 +90,170 @@ class JSQuicSocketListener : public QuicSocketListener {
...
+  void MemoryInfo(MemoryTracker* tracker) const override {
+    tracker->TrackFieldWithSize("data", data_.length());
+    tracker->TrackFieldWithSize("label", strlen(diagnostic_label_));
+  }
+
+  SET_MEMORY_INFO_NAME(QuicPacket);
+  SET_SELF_SIZE(QuicPacket);
+
+ private:
+  MaybeStackBuffer<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> data_;
```
addaleax: This means we always allocate the max packet length, even if the data packet is smaller (which I assume is the common case)… is that okay?

jasnell: Currently, it's not yet clear what the most common case will be. We need to do some testing and see what a reasonable default max is. We can adjust this in the future once we have a better idea

src/node_quic_socket.cc
jasnell: I use this pattern fairly extensively throughout. These can be updates later on in a separate PR

src/node_quic_default_application.cc
jasnell: Can I totally disagree with that? ;-)

jasnell:
Landed!

src/node_quic_default_application.cc
addaleax: You can, but it’s hard to reason about the code locally here if the type isn’t obvious :)

PR #233, opened 2019-12-19
quic: fix compiler error on macos

danbev:
Currently, the following compiler error is generated on macos:
```console
In file included from ../src/node_quic_socket.cc:2:
../src/debug_utils.h:32:27: error:
cannot pass object of non-trivial type 'std::__1::basic_string<char>'
through variadic function; call will abort at runtime [-Wnon-pod-varargs]
  fprintf(stderr, format, std::forward<Args>(args)...);
                          ^
../src/debug_utils.h:48:3: note: in instantiation of function template
specialization 'node::Debug<std::__1::basic_string<char>, int>'
requested here
  Debug(env, cat, format.c_str(), std::forward<Args>(args)...);
  ^
../src/debug_utils.h:78:3: note: in instantiation of function template
specialization 'node::Debug<std::__1::basic_string<char>, int>'
requested here
  Debug(async_wrap->env(),
  ^
../src/debug_utils.h:93:3: note: in instantiation of function template
specialization
'node::UnconditionalAsyncWrapDebug<std::__1::basic_string<char>, int>'
requested here
  UnconditionalAsyncWrapDebug(async_wrap, format,
                              std::forward<Args>(args)...);
  ^
../src/node_quic_socket.cc:418:3: note:
in instantiation of function template specialization
'node::Debug<std::__1::basic_string<char>, int>' requested here
  Debug(this, "Endpoint %s:%d bound",
  ^
1 warning and 1 error generated.
make: *** [node] Error 2
```
This commit passes a const char* to avoid this error.

##### Checklist


jasnell:
Landed

PR #238, opened 2019-12-19
quic: max header pairs and length limit

jasnell:
Options for `maxHeaderPairs` and `maxHeaderLength`


src/node_quic_stream.cc
```diff
@@ -405,9 +405,18 @@ void QuicStream::SetHeadersKind(QuicStreamHeadersKind kind) {
 
 bool QuicStream::AddHeader(std::unique_ptr<QuicHeader> header) {
   Debug(this, "Header Added");
+  size_t len = header->GetLength();
+  auto app = Session()->Application();
```
addaleax:
```suggestion
  QuicApplication* app = Session()->Application();
```

addaleax: Also, side note, our style guide says that getters/setters use `lower_snake_case()` and `set_lower_snake_case()` … currently the QUIC code is doing the exact opposite, and I’ve debating myself whether it’s worth the churn – but ultimately I do think it makes sense to be consistent and align it with the rest of the codebase…

jasnell: I'll go through and update those in a separate PR

jasnell:
Landed

PR #242, opened 2019-12-20
quic: use a utility helper for preferred address

jasnell:
Make working with preferred address a bit easier

### DIFF
--- a/src/node_quic_util-inl.h
+++ b/src/node_quic_util-inl.h
@@ -42,52 +42,6 @@ size_t GetMaxPktLen(const sockaddr* addr) {
       NGTCP2_MAX_PKTLEN_IPV4;
 }
 
-bool ResolvePreferredAddress(
-    Environment* env,
-    int local_address_family,
-    const ngtcp2_preferred_addr* paddr,
-    uv_getaddrinfo_t* req) {
-  int af;
-  const uint8_t* binaddr;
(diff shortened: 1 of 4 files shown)


src/node_quic_util-inl.h
```diff
@@ -234,6 +188,101 @@ const char* QuicError::GetFamilyName() {
...
+}
+const int16_t QuicPreferredAddress::PreferredIPv4Port() const {
+  return paddr_->ipv4_port;
+}
+
+bool QuicPreferredAddress::Use(int family) const {
+  uv_getaddrinfo_t req;
+
+  if (!ResolvePreferredAddress(family, &req) ||
+      req.addrinfo == nullptr) {
```
addaleax: Should `uv_freeaddrinfo()` be called in the second case?

jasnell: No, if `ResolvePreferredAddress()` returns false, then there's no need to call `uv_freeaddrinfo`

addaleax: Yeah, I know, but in the `req.addrinfo == nullptr` case that isn’t quite as obvious?

jasnell: If req.addrinfo is nullptr, there's nothing to free.

jasnell:
Landed

PR #243, opened 2019-12-20
quic: copy the QuicSessionConfig in InitServer

jasnell:
Copy the QuicSessionConfig through assignment in InitServer


jasnell:
Landed!

PR #244, opened 2019-12-20
quic: improve perf of ngtcp2 debug output

jasnell:
Only do work if the category is enabled


jasnell:
Eventually I may do that. I'm leaving it for now because I want to see about potentially emitting trace events here and other places.

jasnell:
Landed!

PR #250, opened 2019-12-29
fix(quic): fix idle timeout unit

oyyd:
The idle_timeout is regarded as milliseconds: https://tools.ietf.org/html/draft-ietf-quic-transport-24#section-18.2

ngtcp2: https://github.com/ngtcp2/ngtcp2/blob/0a29f83a9d2b9435b5b3480c5c14fa505a4026c7/lib/ngtcp2_crypto.c#L448

##### Checklist


addaleax:
Landed in 201625117382f488baa7cfa82af813bbc2a16e04, thanks for fixing this! :)

Issue #249, opened 2019-12-29
QuicClientSession.on('keylog') doesn't work before it's connected


oyyd:
The code below won't emit `keylog` event:

```js
const client = createSocket();

clientSession.on('keylog', (...args) => {
  // this won't get called
});

clientSession.connect({})
```

putting the `clientSession.on('keylog')` inside the callback of `connect()` works: 

```js
const client = createSocket();

clientSession.connect({}, () => {
  clientSession.on('keylog', (...args) => {
    // this won't get called
  });
});
```

From the code here: https://github.com/nodejs/quic/blob/277b68463228676013145080bf81032accff425b/lib/internal/quic/util.js#L584 

`self[kHandle]` is `undefined` before `QuicClientSession` gets connected so that this event won't be emitted from the C++ side.

oyyd:
Just find that it's referenced in #246.

jasnell:
Yep, I'll be opening a PR soon that fixes this

PR #252, opened 2020-01-02
quic: attach keylog and pathvalidation handlers on client init

jasnell:
Enable the keylog and pathValidation event handlers when the
client session handle is created.

Fixes: https://github.com/nodejs/quic/issues/249


jasnell:
Landed

PR #254, opened 2020-01-02
quic: support the NGTCP2_WRITE_STREAM_FLAG_MORE flag

jasnell:
Using NGTCP2_WRITE_STREAM_FLAG_MORE when serializing stream data packets
allows ngtcp2 to pack as much data into a packet as possible, allowing
more efficient transmission when there is data available.


jasnell:
Landed

PR #257, opened 2020-01-02
quic: address path validation todo

jasnell:
There's no need to update the endpoint in the path validation callback

Also fix a minor debug output bug


jasnell:
Landed

PR #258, opened 2020-01-02
quic: emit process warning on --tls-trace

jasnell:


src/quic/node_quic_crypto.cc
```diff
@@ -650,17 +651,24 @@ void SetHostname(SSL* ssl, const std::string& hostname) {
 }  // namespace
 
 void InitializeTLS(QuicSession* session) {
+  static bool warn_trace_tls = true;
```
addaleax: This should not be global mutable state, and stored on the `Environment` like we do for other warnings we only want to emit once

jasnell:
Fixed!

src/quic/node_quic_crypto.cc
```diff
@@ -631,16 +632,22 @@ void SetHostname(SSL* ssl, const std::string& hostname) {
 
 void InitializeTLS(QuicSession* session) {
   QuicCryptoContext* ctx = session->CryptoContext();
-
+  Environment* env = session->env();
   SSL* ssl = ctx->ssl();
   SSL_set_app_data(ssl, session);
   SSL_set_cert_cb(ssl, CertCB, session);
   SSL_set_verify(ssl, SSL_VERIFY_NONE, crypto::VerifyCallback);
+  Environment* env = session->env();
```
addaleax: Is this being added twice now?

```suggestion
```

jasnell:
Landed

PR #259, opened 2020-01-02
quic: use empty string instead of localhost

jasnell:
When servername is an IP address, send an empty string
as SNI rather than localhost


jasnell:
Landed

PR #260, opened 2020-01-03
quic: throw proper errors from c++

jasnell:


src/node_errors.h
```diff
@@ -102,6 +114,18 @@ void PrintErrorString(const char* format, ...);
...
+# define QUIC_PREDEFINED_ERROR_MESSAGES(V)                                   \
+  V(ERR_QUIC_CANNOT_SET_GROUPS, "Cannot set groups")                         \
+  V(ERR_QUIC_FAILURE_SETTING_SNI_CONTEXT, "Failure setting SNI context")
+#else
+# define QUIC_PREDEFINED_ERROR_MESSAGES(V)
+#endif
+
+#define PREDEFINED_ERROR_MESSAGES(V)                                         \
+  BASE_PREDEFINED_ERROR_MESSAGES(V)                                          \
+  QUIC_PREDEFINED_ERROR_MESSAGES(V)
```
addaleax: Tbh, I wouldn’t use conditionals here – everything that happens here happens in inline functions, so if the errors aren’t used, there’s zero overhead.

src/quic/node_quic.cc
```diff
@@ -92,9 +93,10 @@ void QuicInitSecureContext(const FunctionCallbackInfo<Value>& args) {
 
   InitializeSecureContext(sc, side);
 
-  // TODO(@jasnell): Throw a proper node.js error with code
-  if (!crypto::SetGroups(sc, *groups))
-    return env->ThrowError("Failed to set groups");
+  if (!crypto::SetGroups(sc, *groups)) {
+    Isolate* isolate = env->isolate();
+    isolate->ThrowException(ERR_QUIC_CANNOT_SET_GROUPS(isolate));
```
addaleax:
```suggestion
    THROW_ERR_QUIC_CANNOT_SET_GROUPS(env);
```

(afaik that should work)

src/quic/node_quic_session.cc
```diff
@@ -970,8 +972,10 @@ void QuicCryptoContext::OnOCSPDone(
     if (!err) {
       unsigned long err = ERR_get_error();  // NOLINT(runtime/int)
       if (!err) {
-        // TODO(@jasnell): revisit to throw a proper error here
-        return session_->env()->ThrowError("CertCbDone");
+        Isolate* isolate = session_->env()->isolate();
+        isolate->ThrowException(
+            ERR_QUIC_FAILURE_SETTING_SNI_CONTEXT(isolate));
```
addaleax: ditto here, you should be able to use the `THROW_...` variant

jasnell:
Landed

PR #266, opened 2020-01-05
quic: begin cleanup/refactor of QuicBuffer

jasnell:


src/quic/node_quic_session.cc
```diff
@@ -1193,18 +1193,15 @@ void QuicCryptoContext::WriteHandshake(
...
         datalen,
         crypto_level_name(level));
-  // TODO(jasnell): Find a more efficient way of handling this.
-  // The data does have to be copied because it needs to be
-  // kept around until it is ack'd and the given data is not
-  // persistent.
-  MallocedBuffer<uint8_t> buffer(datalen);
-  memcpy(buffer.data, data, datalen);
+  std::unique_ptr<QuicBufferChunk> buffer =
+    std::make_unique<QuicBufferChunk>(datalen);
```
addaleax:
```suggestion
      std::make_unique<QuicBufferChunk>(datalen);
```

src/quic/node_quic_buffer.h
```diff
@@ -9,95 +9,78 @@
...
-                        done_) {
-    data_buf = std::move(buf_);
-  }
+// A QuicBufferChunk contains the actual buffered data
+// along with a callback to be called when the data has
+// been consumed.
+class QuicBufferChunk : public MemoryRetainer {
+ public:
+  // Default non-op done handler.
+  static inline void default_done(int status);
```
addaleax: It doesn’t really make sense for a function that is only used through function pointers to be inline

src/quic/node_quic_buffer.h
```diff
@@ -9,95 +9,78 @@
...
 
-  SET_MEMORY_INFO_NAME(quic_buffer_chunk)
-  SET_SELF_SIZE(quic_buffer_chunk)
+ private:
+  // In practice, the amount of data stored in data_buf_ can vary
+  // broadly from double to quadruple digits. 200 should cover the
+  // most common cases but we can adjust it up or down if necessary
+  // later. Currently, data_buf_ is only used when writing handshake
+  // data. See QuicCryptoContext::WriteHandshake
+  MaybeStackBuffer<uint8_t, 200> data_buf_;
```
addaleax: I think we had this conversation before, but can you add a `TODO` comment here to not use `MaybeStackBuffer`? `QuicBufferChunk` instances are heap-allocated anyway, so the extra data could ideally go into the same heap allocation, and if there is no extra data then using `MaybeStackBuffer` adds 200 bytes of overhead per chunk

src/quic/node_quic_buffer.cc
```diff
@@ -0,0 +1,169 @@
...
+    // Otherwise, decrement the amt and advance the read head
+    // one space and iterate from there.
+    amt -= len;
+    rlength_ -= len;
+    head_ = head_->next_.get();
+  }
+}
+
+size_t QuicBuffer::DrainInto(
+    add_fn add_to_list,
```
addaleax: Any reason to switch away from the template variant? The function isn’t stored anywhere, so using `std::function` will have unnecessary overhead, 1) in terms of an extra temporary heap allocation and, more importantly, 2) because the compiler can’t inline it anymore

src/quic/node_quic_buffer-inl.h
```diff
@@ -0,0 +1,120 @@
...
+}
+
+void QuicBufferChunk::Done(int status) {
+  if (done_called_) return;
+  done_called_ = true;
+  if (done_ != nullptr)
+    std::move(done_)(status);
+}
+
+void QuicBufferChunk::MemoryInfo(MemoryTracker* tracker) const {
```
addaleax: Fwiw, this can also go into the `.cc` file, it’s never used as an inline function, and it removes the need to include `memory_tracker-inl.h` when including this file

src/quic/node_quic_buffer.h
jasnell: I'll just make the change directly in this PR :-)

jasnell: (and thanks for the reminder that we did discuss that before :-D ..)

src/quic/node_quic_buffer.cc
jasnell: Compiler errors from the calls to it from the other inline DrainInto variants (unresolved symbol). I'll tweak it tho.

jasnell:
Landed

PR #267, opened 2020-01-05
quic: various cleanups in quic_crypto

jasnell:
Mainly: using arrays directly instead of std::array where appropriate


jasnell:
Landed

PR #269, opened 2020-01-06
quic: inline and const where possible

jasnell:
@addaleax, can I ask you to given this a good look through to make sure the various inlines make sense? :-)

### DIFF
--- a/src/quic/node_quic_http3_application.h
+++ b/src/quic/node_quic_http3_application.h
@@ -133,7 +133,7 @@ class Http3Application final :
   void MemoryInfo(MemoryTracker* tracker) const override;
 
  private:
-  nghttp3_conn* Connection() { return connection_.get(); }
+  nghttp3_conn* Connection() const { return connection_.get(); }
   QuicStream* FindOrCreateStream(int64_t stream_id);
 
   bool CreateAndBindControlStream();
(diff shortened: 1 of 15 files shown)


src/quic/node_quic_session.h
```diff
@@ -222,45 +222,47 @@ enum QuicSessionStatsIdx : int {
 
 class QuicSessionListener {
  public:
-  virtual ~QuicSessionListener();
+  virtual inline ~QuicSessionListener();
```
addaleax: So … `virtual inline` rarely makes sense, unfortunately. It only helps when the compiler can be *sure* of the actual subclass type for a given object in order to know which function to inline, but that doesn’t apply here, because we call these methods through a pointer that could ultimately refer to any given subclass, as far as I can tell – and in that case it reduces complexity a bit to keep it non-inline.

src/quic/node_quic_http3_application.h
```diff
@@ -133,7 +133,7 @@ class Http3Application final :
   void MemoryInfo(MemoryTracker* tracker) const override;
 
  private:
-  nghttp3_conn* Connection() { return connection_.get(); }
+  nghttp3_conn* Connection() const { return connection_.get(); }
```
addaleax: Not sure, but maybe this is a good PR to also align the getter naming with our style guide and use `connection()`, `session()`, etc.?

jasnell: yeah, but let me land the other opens ones first, then I'll rebase and make that change

jasnell:
Landed

PR #275, opened 2020-01-09
quic: refactoring SendPendingData

jasnell:
This is a multi-step cleanup of `QuicApplication::SendPendingData`, allowing both `DefaultApplication` and `Http3Application` to use the same base `SendPendingData()` logic and making the iteration over `QuicStream` instances in `DefaultApplication` more efficient by only iterating over those that have data to send.

### DIFF
--- a/src/quic/node_quic_session-inl.h
+++ b/src/quic/node_quic_session-inl.h
@@ -149,7 +149,7 @@ ssize_t QuicApplication::WriteVStream(
     uint8_t* buf,
     ssize_t* ndatalen,
     const StreamData& stream_data) {
-  CHECK_LE(stream_data.count, 16);
+  CHECK_LE(stream_data.count, MAX_VECTOR_COUNT);
   return ngtcp2_conn_writev_stream(
     session()->connection(),
     &path->path,
--- a/src/quic/node_quic_session-inl.h
+++ b/src/quic/node_quic_session-inl.h
@@ -138,6 +138,35 @@ bool QuicCryptoContext::SetupInitialKey(const ngtcp2_cid* dcid) {
 
 QuicApplication::QuicApplication(QuicSession* session) : session_(session) {}
 
+void QuicApplication::set_stream_fin(int64_t stream_id) {
+  QuicStream* stream = session()->FindStream(stream_id);
+  if (stream != nullptr)
+    stream->set_fin_sent();
+}
+
+ssize_t QuicApplication::WriteVStream(
+    QuicPathStorage* path,
+    uint8_t* buf,
+    ssize_t* ndatalen,
+    const StreamData& stream_data) {
+  CHECK_LE(stream_data.count, 16);
+  return ngtcp2_conn_writev_stream(
+    session()->connection(),
+    &path->path,
+    buf,
+    session()->max_packet_length(),
+    ndatalen,
+    stream_data.remaining > 0 ?
+        NGTCP2_WRITE_STREAM_FLAG_MORE :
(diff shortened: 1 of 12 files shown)


src/quic/node_quic_session-inl.h
```diff
@@ -138,6 +138,35 @@ bool QuicCryptoContext::SetupInitialKey(const ngtcp2_cid* dcid) {
...
+  if (stream != nullptr)
+    stream->set_fin_sent();
+}
+
+ssize_t QuicApplication::WriteVStream(
+    QuicPathStorage* path,
+    uint8_t* buf,
+    ssize_t* ndatalen,
+    const StreamData& stream_data) {
+  CHECK_LE(stream_data.count, 16);
```
addaleax: Should this be `MAX_VECTOR_COUNT`?

src/quic/node_quic_session.cc
```diff
@@ -1098,6 +1098,121 @@ void QuicCryptoContext::WriteHandshake(
...
+  QuicStream* stream = session()->FindStream(stream_id);
+  if (stream != nullptr) {
+    stream->Acknowledge(offset, datalen);
+    ResumeStream(stream_id);
+  }
+}
+
+bool QuicApplication::SendPendingData() {
+  // The maximum number of packets to send per call
+  static constexpr size_t MAX_PACKETS = 16;
```
addaleax: `kMaxPackets`?

src/quic/node_quic_stream.h
```diff
@@ -323,118 +287,83 @@ class QuicStream : public AsyncWrap, public StreamBase {
...
+
+  // Submits trailing headers. Returns false if headers are not
+  // supported on the underlying QuicApplication.
+  inline bool SubmitTrailers(v8::Local<v8::Array> headers);
+
+  // Required for StreamBase
+  inline bool IsAlive() override;
+  inline bool IsClosing() override;
+  inline int ReadStart() override;
+  inline int ReadStop() override;
```
addaleax: These are virtual, so making them inline doesn’t make much sense

src/quic/node_quic_stream.cc
```diff
@@ -139,21 +193,18 @@ void QuicStream::Destroy() {
...
+  if (is_writable()) {
+    Debug(this, "Shutdown writable side");
+    stream_stats_.closing_at = uv_hrtime();
+    set_write_close();
+    session()->ResumeStream(stream_id_);
+  }
 
-  return 1;
+  req_wrap->Done(0);
+  return 0;
```
addaleax: Why this? `return 1;` is exactly the right thing to indicate synchronous success:

https://github.com/nodejs/quic/blob/399cc9cbd9ec7cbe9805f5af45a35d097489a9a0/src/stream_base.h#L218-L221

jasnell: Arg... I keep mixing it up with DoWrite, which has return 0 for success.

addaleax: Inspired by this: https://github.com/nodejs/node/pull/31283

I think we can align the behaviour with `DoWrite()` here if you like, it’s just that we currently don’t have any `StreamBase`s that finish writes fully synchronously afaik (and if they do, it’s through `DoTryWrite()` rather than `DoWrite()`…)

jasnell: Eventually, but I plan to explore moving away from StreamBase entirely before too long to see if a model like Jeremiah's BOB streams would be more efficient. I'll leave it like this for now.

jasnell:
Landed

Issue #209, opened 2019-12-03
TODO: Support HTTP/3 Push Streams


jasnell:
Push streams have not yet been implemented in the Http3 Application

jasnell:
Fixed by: https://github.com/nodejs/quic/pull/280

PR #280, opened 2020-01-13
quic: starting http/3 push streams support

jasnell:
Fixes: https://github.com/nodejs/quic/issues/209

This is a beginning of push stream support. There are a few API nits we need to figure out.

Specifically, a push stream is *always* associated with a client-initiated bidirectional stream. Currently, however, when a push stream is received, it is emitted using the `on('stream')` event on the `QuicClientSession`. When received, the pushed stream is currently not associated with the original client-initiated bidirectional stream. The pushed request headers, however, are emitted using the `pushHeaders` event on the original stream.

```js
  const req = client.connect({ /* ... */ });

  req.on('stream', (stream) => {
    // this is the actual pushed stream
    stream.on('initialHeaders', (headers) => { /** ... **/ });
    stream.resume();
  });

  req.on('secure', (servername, alpn, cipher) => {
    const stream = req.openStream();

    stream.on('pushHeaders', (headers, push_id) => {
      // push promise headers
    });

    assert(stream.submitInitialHeaders({
      ':method': 'POST',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/',
    }));

    stream.end('hello world');
    stream.resume();
  }));
```

This is obviously not very ergonomic.

The other bit is that for push streams there are always two identifiers... a push ID and the stream ID. When the `'pushHeaders'` event is emitted, all we know is the push ID, we won't know which stream ID fulfills the push ID until the push stream is actually created.

The final bit is that QUIC provides a way of cancelling / rejecting a push stream but I'm not yet exposing that. What I'm thinking is passing a callback function when we receive the initial push promise -- user code can call that function to reject the push promise.

This PR just starts getting the basic mechanism in place but it will need to be refactored.

What I'm thinking is a refactor that emits `'on('pushPromise')` and `'on('pushStream')` on the `QuicStream` object:

```js
  const req = client.connect({ /* ... */ });

  req.on('stream', (stream) => {
    // this is the actual pushed stream
    stream.on('initialHeaders', (headers) => { /** ... **/ });
    stream.resume();
  });

  req.on('secure', (servername, alpn, cipher) => {
    const stream = req.openStream();

    stream.on('pushPromise', (push_id, headers, reject) => {
      // push promise headers
      // To reject the push promise, call reject();
    });

    stream.on('pushStream', (push_id, pushStream) => {
      pushStream.on('initialHeaders', (headers) => { /** ... **/ });
      pushStream.resume();
    }}

    assert(stream.submitInitialHeaders({
      ':method': 'POST',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/',
    }));

    stream.end('hello world');
    stream.resume();
  }));
```

Before I start to implement the refactored API, however, I'd like to get some feedback (/cc @mcollina @addaleax)


jasnell:
Note: needs to be rebased

test/parallel/test-quic-http3-push.js
```diff
@@ -0,0 +1,145 @@
...
+
+    const push = stream.pushStream({
+      ':method': 'GET',
+      ':scheme': 'https',
+      ':authority': 'localhost',
+      ':path': '/foo'
+    });
+    assert(push);
+    push.submitInitialHeaders({ ':status': '200' });
+    // TODO(@jasnell): There's currently a bug if we only call end() to write.
```
addaleax: Just curious, what happens in that case?

jasnell: At the time there was a failed assertion. I need to retest.

jasnell:
Landed. Will be investigating the TODO comments in the next round.

PR #281, opened 2020-01-14
quic: multiple code cleanups

jasnell:
Multiple cleanups that have been on the todo for a while

### DIFF
--- a/src/node_sockaddr-inl.h
+++ b/src/node_sockaddr-inl.h
@@ -69,8 +69,8 @@ bool SocketAddress::is_numeric_host(const char* hostname) {
 }
 
 bool SocketAddress::is_numeric_host(const char* hostname, int family) {
-  std::array<uint8_t, sizeof(struct in6_addr)> dst;
-  return inet_pton(family, hostname, dst.data()) == 1;
+  in6_addr dst;
+  return inet_pton(family, hostname, &dst) == 1;
 }
 
 sockaddr_storage* SocketAddress::ToSockAddr(
--- a/src/quic/node_quic_buffer.h
+++ b/src/quic/node_quic_buffer.h
@@ -17,7 +17,7 @@ namespace quic {
 
 class QuicBuffer;
 
-constexpr size_t MAX_VECTOR_COUNT = 16;
+constexpr size_t kMaxVectorCount = 16;
 
 typedef std::function<void(int status)> done_cb;
 
@@ -148,7 +148,7 @@ class QuicBuffer : public MemoryRetainer {
   inline size_t DrainInto(
       std::vector<uv_buf_t>* list,
       size_t* length = nullptr,
-      size_t max_count = MAX_VECTOR_COUNT);
+      size_t max_count = kMaxVectorCount);
 
   template <typename T>
   inline size_t DrainInto(
--- a/src/quic/node_quic_crypto.cc
+++ b/src/quic/node_quic_crypto.cc
@@ -35,23 +35,24 @@ using v8::Value;
 
 namespace quic {
 
-constexpr int NGTCP2_CRYPTO_SECRETLEN = 64;
-constexpr int NGTCP2_CRYPTO_KEYLEN = 64;
-constexpr int NGTCP2_CRYPTO_IVLEN = 64;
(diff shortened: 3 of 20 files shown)


src/quic/node_quic_crypto.cc
```diff
@@ -52,6 +52,7 @@ constexpr char QUIC_SERVER_HANDSHAKE_TRAFFIC_SECRET[] =
     "QUIC_SERVER_HANDSHAKE_TRAFFIC_SECRET";
 constexpr char QUIC_SERVER_TRAFFIC_SECRET_0[] =
     "QUIC_SERVER_TRAFFIC_SECRET_0";
+}
```
danbev: Nit: This needs a closing namespace comment:
```console
src/quic/node_quic_crypto.cc:55:  Anonymous namespace should be terminated with "// namespace"  [readability/namespace] [5]
```

src/node_sockaddr-inl.h
```diff
@@ -69,8 +69,8 @@ bool SocketAddress::is_numeric_host(const char* hostname) {
 }
 
 bool SocketAddress::is_numeric_host(const char* hostname, int family) {
-  std::array<uint8_t, sizeof(struct in6_addr)> dst;
-  return inet_pton(family, hostname, dst.data()) == 1;
+  uint8_t dst[sizeof(struct in6_addr)];
```
addaleax: Or just `in6_addr dst;` with `&dst` on the next line?

src/quic/node_quic_buffer.h
```diff
@@ -116,7 +116,7 @@ class QuicBuffer : public MemoryRetainer {
 
   ~QuicBuffer() override {
     Cancel();  // Cancel the remaining data
-    CHECK_EQ(length_, 0);
+    DCHECK_EQ(length_, 0);
```
addaleax: I’d prefer to keep this as is, at least as long as the QUIC code is under active development

src/quic/node_quic_crypto.cc
```diff
@@ -80,24 +81,24 @@ bool DeriveTokenKey(
...
-  std::array<uint8_t, 32> md;
+  uint8_t rand[16];
+  uint8_t md[32];
 
   const EVP_MD* meth = EVP_sha256();
   unsigned int mdlen = EVP_MD_size(meth);
   DeleteFnPtr<EVP_MD_CTX, EVP_MD_CTX_free> ctx;
   ctx.reset(EVP_MD_CTX_new());
-  CHECK(ctx);
+  DCHECK(ctx);
```
addaleax: Ditto here and elsewhere – if this condition is ever false, there’s a crash that’s a lot harder to debug without the `CHECK` message. `CHECK`s are ultimately very cheap (branch prediction is basically always going to do the right things for them), especially if they only perform single-word comparisons; I’d use `DCHECK`s only when there’s real performance overhead associated with the condition itself or in really hot paths (after benchmarking).

jasnell:
@addaleax ... please take another look :-) ... there are more cleanup chores to do but let's get this one landed first.

src/quic/node_quic_session.cc
```diff
@@ -1367,35 +1367,35 @@ QuicSession::QuicSession(
 
   session_stats_.created_at = uv_hrtime();
 
-  if (wrap->DefineOwnProperty(
+  CHECK(wrap->DefineOwnProperty(
           env()->context(),
           env()->state_string(),
           state_.GetJSArray(),
-          PropertyAttribute::ReadOnly).IsNothing()) return;
+          PropertyAttribute::ReadOnly).IsJust());
```
addaleax: I prefer handling errors, but if you do this, there is a `.Check()` method on `Maybe<...>` that we tend to prefer in our codebase over `CHECK((...).IsJust());` (it does the same thing)

jasnell: Before it moves out of experimental, we'll handle these. For now while we're actively developing, I want to make them checks because they really shouldn't fail but I want to make sure while we're going through building tests. I'll add a todo note to handle them.

jasnell: (note, they weren't being properly handled previously)

jasnell:
Landed

PR #286, opened 2020-01-15
quic: QuicCID refactors and docs

jasnell:
There are two commits here:

1. Makes some simple fixes and adds some docs
2. The main commit, uses QuicCID more consistently

### DIFF
--- a/src/quic/node_quic_socket.cc
+++ b/src/quic/node_quic_socket.cc
@@ -658,7 +658,7 @@ bool QuicSocket::SendStatelessReset(
   if (pktlen < kMinStatelessResetLen)
     return false;
 
-  GenerateResetToken(token, reset_token_secret_, cid.cid());
+  GenerateResetToken(token, reset_token_secret_, cid);
   EntropySource(random, kRandlen);
 
   auto packet = QuicPacket::Create("stateless reset", pktlen);
@@ -707,14 +707,8 @@ bool QuicSocket::SendRetry(
 
   // Retry tokens are generated cryptographically. They
   // aren't super expensive but they are still not zero-cost.
-  if (!GenerateRetryToken(
-          token,
-          &tokenlen,
-          remote_addr,
-          dcid.cid(),
-          token_secret_)) {
+  if (!GenerateRetryToken(token, &tokenlen, remote_addr, dcid, token_secret_))
     return false;
-  }
 
   ngtcp2_pkt_hd hd;
   hd.version = version;
@@ -724,7 +718,7 @@ bool QuicSocket::SendRetry(
   hd.token = nullptr;
   hd.tokenlen = 0;
   hd.len = 0;
-  hd.dcid = *scid.cid();
+  hd.dcid = *scid;
   hd.scid.datalen = kScidLen;
 
   EntropySource(hd.scid.data, kScidLen);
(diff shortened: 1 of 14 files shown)


src/quic/node_quic_socket.cc
```diff
@@ -457,25 +484,30 @@ void QuicSocket::OnReceive(
   QuicCID dcid(pdcid, pdcidlen);
   QuicCID scid(pscid, pscidlen);
 
+  // TODO(@jasnell): It would be fantastic if Debug() could be
+  // modified to accept objects with a ToString-like capability
+  // similar to what we can do with TraceEvents... that would
+  // allow us to pass the QuicCID directly to Debug and have it
+  // converted to hex only if the category is enabled so we can
+  // skip committing resources here.
```
addaleax: Fully agree :+1: 

It’s a bit tricky because that means we can’t use `fprintf()` directly anymore, but that’s probably pretty okay actually given that having a C++-y replacement would also help solve issues like https://github.com/nodejs/node/issues/28761.

src/quic/node_quic_socket.cc
```diff
@@ -509,10 +541,26 @@ void QuicSocket::OnReceive(
...
-    // or (in the future) a CONNECTION_CLOSE packet.
+    // or a CONNECTION_CLOSE packet.
     if (!session) {
       Debug(this, "Unable to create a new server QuicSession.");
-
+      // If the packet contained a short header, we might need to send
+      // a stateless reset. The stateless reset contains a token derived
+      // from the received destination connection ID.
+      //
+      // TODO(@jasnell): Stateless resets are generated programatically
```
addaleax:
```suggestion
      // TODO(@jasnell): Stateless resets are generated programmatically
```

src/quic/node_quic_socket.cc
```diff
@@ -669,6 +742,10 @@ bool QuicSocket::SendRetry(
   return SendPacket(local_addr, remote_address, std::move(packet)) == 0;
 }
 
+// Inspects the packet and possible accepts it as a new
```
addaleax:
```suggestion
// Inspects the packet and possibly accepts it as a new
```

jasnell:
btw, QuicCID currently isn't as efficient as it could be in the common case. I'll be working on that today

jasnell:
@addaleax ... added a new commit that avoids copying in QuicCID when not necessary to do so. Can you please take a look when you get a moment?

addaleax:
@jasnell Did you benchmark this? (I can also do that if you like.) My best intuition would be that the new code actually adds non-trivial overhead without much gain.

Copying an `ngtcp2_cid` is very cheap, on both x64 and arm it compiles down to a mere 4 instructions (https://godbolt.org/z/uhTCFQ), whereas using `std::vector<...>` comes with the overhead of an extra heap allocation, and it requires updating at least the same amount of memory for a copy or move operation (`std::vector<>` is usually 3 word-sized values itself, i.e. the management memory for a vector is about as large as a `ntcp2_cid` itself).

jasnell:
No, didn't benchmark it yet but thinking it you're right. We're generally talking only 32 bytes. I'll drop the additional commit but keep it around in a separate branch just in case we want to move over for some reason

jasnell:
Ok, take another look and let me know if that's better :-)

jasnell:
Landed!

PR #288, opened 2020-01-16
quic: refactor QuicBuffer/QuicStream, fix http3 push

jasnell:
Working through some issues with the Http3Application, specifically with push but the issue likely effects more than that and is only triggered with push streams.

Specifically, 

Given the following code running on the server inside the `session.on('stream')` event, the client receives the data `testing` as expected and the stream on the client side closes normally as expected.

```js
    const push = stream.pushStream({
      ':method': 'GET',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/foo'
    });
    assert(push);
    push.submitInitialHeaders({ ':status': '200' });
    push.write('test');
    push.end('ing');
    push.on('close', common.mustCall());
    push.on('finish', common.mustCall());
```

However, if we switch to either of the two following, the data is delivered to the client side but the close, headers, and data are delivered out of order causing issues and stopping the push stream from closing normally on the client.

```js
    const push = stream.pushStream({
      ':method': 'GET',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/foo'
    });
    assert(push);
    push.submitInitialHeaders({ ':status': '200' });
    push.write('');
    push.end('testing');
    push.on('close', common.mustCall());
    push.on('finish', common.mustCall());
```

```js
    const push = stream.pushStream({
      ':method': 'GET',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/foo'
    });
    assert(push);
    push.submitInitialHeaders({ ':status': '200' });
    push.end('testing');
    push.on('close', common.mustCall());
    push.on('finish', common.mustCall());
```

Wrapping the `push.end('testing')` in a `setImmediate()`, however, causes the code to work.

```js
    const push = stream.pushStream({
      ':method': 'GET',
      ':scheme': 'https',
      ':authority': 'localhost',
      ':path': '/foo'
    });
    assert(push);
    push.submitInitialHeaders({ ':status': '200' });
    setImmediate(() => push.end('testing'));
    push.on('close', common.mustCall());
    push.on('finish', common.mustCall());
```

To complicate things, writing only with `end('hello world')` on the bidirectional request/response stream appears to work perfectly, so the issue *appears* to be limited to the unidirectional push streams.

On the server side, it *appears* that the data for the push stream is being serialized correctly to a QUIC packet and sent to the client. A QUIC packet with an appropriate fin flag is also sent. The client QuicSession also successfully receives the data and the fin. The *challenge*, however, appears to be the fact that in the push promise and response headers are delivered on a *different* QUIC stream and there are timing differences between when the *data* on the push stream is sent and when the *headers* for that stream are sent and processed, causing some weird processing issue on the client side.

It's *POSSIBLE* that we're seeing a bug in nghttp3 with regards to unidirectional push streams -- there was another bug that I uncovered a couple days ago in nghttp3 in relation to unidirectional streams, but it's really not yet clear.

### DIFF
--- a/src/node_bob.h
+++ b/src/node_bob.h
@@ -1,0 +1,109 @@
+#ifndef SRC_NODE_BOB_H_
+#define SRC_NODE_BOB_H_
+
+#include <functional>
+
+namespace node {
+namespace bob {
+
+constexpr size_t kMaxCountHint = 16;
+
+// Negative status codes indicate error conditions.
+enum Status : int {
+  // Indicates that an attempt was made to pull after end.
+  STATUS_EOS = -1,
+
+  // Indicates the end of the stream. No additional
+  // data will be available and the consumer should stop
+  // pulling.
+  STATUS_END = 0,
+
+  // Indicates that there is additional data available
+  // and the consumer may continue to pull.
+  STATUS_CONTINUE = 1,
+
+  // Indicates that there is no additional data available
+  // but the stream has not ended. The consumer should not
+  // continue to pull but may resume pulling later when
+  // data is available.
+  STATUS_BLOCK = 2,
+
+  // Indicates that there is no additional data available
+  // but the stream has not ended and the source will call
+  // next again later when data is available. STATUS_WAIT
+  // must not be used with the OPTIONS_SYNC option.
+  STATUS_WAIT = 3,
+};
+
+enum Options : int {
+  OPTIONS_NONE = 0,
+
+  // Indicates that the consumer is requesting the end
+  // of the stream.
+  OPTIONS_END = 1,
+
+  // Indicates that the consumer requires the source to
+  // invoke Next synchronously. If the source is
+  // unable to provide data immediately but the
+  // stream has not yet ended, it should call Next
+  // using STATUS_BLOCK. When not set, the source
+  // may call Next asynchronously.
+  OPTIONS_SYNC = 2
+};
+
+// There are Sources and there are Consumers.
+//
+// Consumers get data by calling Source::Pull,
+// optionally passing along a status and allocated
+// buffer space for the source to fill, and a Next
+// function the Source will call when data is
+// available.
+//
+// The source calls Next to deliver the data. It can
+// choose to either use the allocated buffer space
+// provided by the consumer or it can allocate its own
+// buffers and push those instead. If it allocates
+// its own, it can send a Done function that the
+// Sink will call when it is done consuming the data.
+template <typename T>
+class Source {
+ public:
+  using Done = std::function<void(size_t)>;
+  using Next = std::function<void(int, const T*, size_t count, Done done)>;
+
+  virtual int Pull(
+      Next next,
+      int options,
+      T* data,
+      size_t count,
+      size_t max_count_hint = kMaxCountHint) = 0;
+};
+
+template <typename T>
+class SourceImpl : public Source<T> {
+ public:
+  int Pull(
+      Next next,
+      int options,
+      T* data,
+      size_t count,
+      size_t max_count_hint = kMaxCountHint) override;
+
+  bool is_eos() const { return eos_; }
+
+ protected:
+  virtual int DoPull(
+      Next next,
+      int options,
+      T* data,
+      size_t count,
+      size_t max_count_hint) = 0;
+
+ private:
+  bool eos_ = false;
+};
+
+}  // namespace bob
+}  // namespace node
+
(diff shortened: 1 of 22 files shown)


jasnell:
The first commit in this PR is a handful of cleanups and docs. I'm still working on identifying the fix necessary.

src/quic/node_quic_http3_application.cc
```diff
@@ -132,44 +110,87 @@ void Http3Header::MemoryInfo(MemoryTracker* tracker) const {
...
+// The push promise is sent on the request/response bidirectional
+// stream.
+// The push fulfillment is a unidirectional stream opened by the
+// server that contains the push id, the response header block, and
+// the response payload.
+// Here's where it can get a bit complicated: the server sends the
+// push promise and the push fulfillment on two different, and
+// independent QUIC streams. The push id is used to correlate
+// those on the client side, but, it's entirely possible for the
+// client to receive the push fullment before it actually receives
```
addaleax:
```suggestion
// client to receive the push fulfillment before it actually receives
```

src/quic/node_quic_http3_application.cc
```diff
@@ -132,44 +110,87 @@ void Http3Header::MemoryInfo(MemoryTracker* tracker) const {
...
+// the push promise. It's *unlikely*, but it's possible. Fortunately,
+// nghttp3 handles the complexity of that for us internally but
+// makes for some weird timing and could lead to some amount of
+// buffering to occur.
+//
+// The *logical* order of events from the client side *should*
+// be: (a) receive the push promise containing assumed request
+// headers, (b) receiver the push fulfillment containing the
```
addaleax:
```suggestion
// headers, (b) receive the push fulfillment containing the
```

jasnell:
@addaleax .. this is ready. The issue with the http3 push stuff was largely that the signal to end the stream was not triggering correctly, which was largely caused by a deficiency in how QuicBuffer and QuicStream were tracking state. I took it as an opportunity to do some refactoring and pull in a modified form of @Fishrock123's bob stream interface as I've been wanting to do for a while now. The implementation is much cleaner over the previous DrainInto approach and the issue with http3 push should be resolved.

jasnell:
@Fishrock123 ... the node_bob.h API in this is a bit different from what you've spec'd. Same fundamental model but modified to fit more of what I needed here. We can continue to iterate and tweak on this but definition worth taking a look through.

src/node_bob.h
```diff
@@ -0,0 +1,112 @@
...
+  // invoke Next synchronously. If the source is
+  // unable to provide data immediately but the
+  // stream has not yet ended, it should call Next
+  // using STATUS_BLOCK. When not set, the source
+  // may call Next asynchronously.
+  OPTIONS_SYNC = 2
+};
+
+
+// The node::bob protocol is fairly straight forward:
```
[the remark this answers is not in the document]
jasnell: > (That being said, I can’t really agree with this statement :))

;-) ... I'll just take that bit out then :-)

jasnell:
> doesn’t really look much better or worse to me

Better in the sense that (a) there's more consistency in data read between DefaultApplication and Http3Application, (b) end-of-stream is now more inline which is what actually solved the issue with http3 push, and (c) DrainInto was really convoluted before.

jasnell:
With regards to QuicBuffer, yes, it's a confusing object. I plan to add some additional documentation to it to explain it a bit more.

src/quic/node_quic_buffer.cc
```diff
@@ -16,27 +16,27 @@ void QuicBufferChunk::MemoryInfo(MemoryTracker* tracker) const {
   tracker->TrackField("next", next_);
 }
 
-size_t QuicBuffer::push(uv_buf_t* bufs, size_t nbufs, DoneCB done) {
+size_t QuicBuffer::Push(uv_buf_t* bufs, size_t nbufs, DoneCB done) {
   size_t len = 0;
   if (nbufs == 0 || bufs == nullptr || is_empty(bufs[0])) {
-    done(0);
+    Done(0);
```
addaleax: This would call the wrong function now, wouldn’t it?

jasnell: good catch, bad search/replace.

jasnell:
Looks like this generally fixes the issue but there's still some flakiness in the push test that I need to chase down... some odd timing issue.

jasnell:
Landed

PR #294, opened 2020-01-22
quic: additional cleanups and fixups

jasnell:


src/quic/node_quic_session.h
```diff
@@ -1437,6 +1423,15 @@ class QuicSession : public AsyncWrap,
 
   static const ngtcp2_conn_callbacks callbacks[2];
 
+  class StatsDebug {
+   public:
+    StatsDebug(QuicSession* session) : session_(session) {}
+    std::string ToString();
```
addaleax:
```suggestion
    std::string ToString() const;
    
```

(This is required by https://github.com/nodejs/node/pull/31446. The blank line before `public:`, `private:` etc. is something that the linter should complain about?)

src/quic/node_quic_socket.cc
```diff
@@ -658,15 +649,15 @@ bool QuicSocket::SendStatelessReset(
...
+  StatelessResetToken token(reset_token_secret_, cid);
   EntropySource(random, kRandlen);
 
   auto packet = QuicPacket::Create("stateless reset", pktlen);
   ssize_t nwrite =
       ngtcp2_pkt_write_stateless_reset(
         reinterpret_cast<uint8_t*>(packet->data()),
         NGTCP2_MAX_PKTLEN_IPV4,
-        token,
+        const_cast<uint8_t*>(token.data()),
```
addaleax: This is an ngtctp2 bug, right? Afaict `ngtcp2_pkt_write_stateless_reset()` should use `const uint8_t*` for `stateless_reset_token` and `rand`?

src/quic/node_quic_socket.h
```diff
@@ -531,6 +531,14 @@ class QuicSocket : public AsyncWrap,
 
   SendWrap* last_created_send_wrap_ = nullptr;
 
+  class StatsDebug {
+   public:
+    StatsDebug(QuicSocket* socket) : socket_(socket) {}
+    std::string ToString();
```
addaleax:
```suggestion
    std::string ToString() const;

```

src/quic/node_quic_stream.h
```diff
@@ -80,6 +83,8 @@ class QuicHeader : public MemoryRetainer {
   // Returns the total length of the header in bytes
   // (including the name and value)
   virtual size_t length() const = 0;
+
+  inline std::string ToString();
```
addaleax:
```suggestion
  inline std::string ToString() const;
```

src/quic/node_quic_stream.h
```diff
@@ -371,6 +381,14 @@ class QuicStream : public AsyncWrap,
 
   ListNode<QuicStream> stream_queue_;
 
+  class StatsDebug {
+   public:
+    StatsDebug(QuicStream* stream) : stream_(stream) {}
+    std::string ToString();
```
addaleax:
```suggestion
    std::string ToString() const;
    
```

src/quic/node_quic_stream.h
```diff
@@ -371,6 +381,14 @@ class QuicStream : public AsyncWrap,
 
   ListNode<QuicStream> stream_queue_;
 
+  class StatsDebug {
+   public:
+    StatsDebug(QuicStream* stream) : stream_(stream) {}
```
addaleax:
```suggestion
    explicit StatsDebug(QuicStream* stream) : stream_(stream) {}
```

(Also for linting)

src/quic/node_quic_socket.cc
jasnell: I believe so yes

addaleax: https://github.com/ngtcp2/ngtcp2/pull/210

src/quic/node_quic_util.h
```diff
@@ -87,6 +87,19 @@ class StatsBase {
...
+
+  using AddField = std::function<void(const char*, uint64_t)>;
+
+  // The StatsDebug utility is used when StatsBase is destroyed
+  // to output statistical information.
+  struct StatsDebug {
+    Q* ptr;
+    explicit StatsDebug(Q* ptr_) : ptr(ptr_) {}
+    std::string ToString() const;
+    void ToString(AddField add_field) const;
```
addaleax: I realize it’s not that important for debug-only code, but typically, this would be implemented by making function argument a template (in order to avoid the overhead of `std::function` here)

src/quic/node_quic_util.h
```diff
@@ -73,7 +73,7 @@ enum QuicErrorFamily : int32_t {
 
 // StatsBase is a utility help for classes (like QuicSession)
 // that record performance statistics
-template <typename T>
+template <typename T, typename Q>
```
addaleax: `Q` is currently always an `AsyncWrap` subclass, and that’s the only property of it that is being used – maybe this doesn’t need to be a template parameter?

src/quic/node_quic_util.h
jasnell: Not sure I follow

addaleax: I’m taking about something like 

```c++
    template <typename Fn>
    void ToString(Fn&& add_field);
```

which is otherwise the same but doesn’t use an `std::function` to pass the function pointer.

Like I said, this isn’t particularly important for debug-only code, so feel free to ignore.

src/quic/node_quic_util.h
jasnell: Thought of that also, but the call to Debug() in ~StatsDebug doesn't like passing either `this` or `reinterpret_cast<AsyncWrap*>` as the first argument. I'm liking missing something simple.

addaleax: Hm yeah… I hadn’t thought of that. CRTP is the typical approach for that – I think we could get rid of the double templating by inferring `T` from `Q`, similar to the traits approach used for the header refcount classes, but again, nothing important that needs to be addressed now

src/quic/node_quic_util.h
jasnell: Got it... made some tweaks. Take a look now :-)

src/quic/node_quic_util.h
jasnell: Went ahead and tweaked it. Using a Traits approach now. Still likely could use some tweak but hopefully much better.

jasnell:
Landed

PR #299, opened 2020-01-23
quic: implement maxConnection limit

jasnell:


src/quic/node_quic_socket.cc
```diff
@@ -1014,12 +1027,13 @@ void NewQuicSocket(const FunctionCallbackInfo<Value>& args) {
...
       retry_token_expiration,
+      max_connections,
       max_connections_per_host,
       max_stateless_resets_per_host,
       options,
-      args[4]->IsTrue() ? QlogMode::kEnabled : QlogMode::kDisabled,
+      args[5]->IsTrue() ? QlogMode::kEnabled : QlogMode::kDisabled,
       session_reset_secret,
-      args[5]->IsTrue());
+      args[7]->IsTrue());
```
addaleax:
```suggestion
      args[6]->IsTrue());
```

(I assume?)

jasnell: I really think it's supposed to be args[7] here.

addaleax: Ah, missed that it was wrong before – makes sense then :)

jasnell: I'm going to be cleaning up those arguments soon, btw... since the number of args keeps growing and most are numeric.

jasnell:
Landed

jasnell:
Ugh... looks like there was a boo boo in this that's causing a segfault. Finding it now

jasnell:
Fixed

PR #300, opened 2020-01-23
src: initial support for setting ipv6 flow label

jasnell:
QUIC recommends setting the IPv6 flow label in various cases.

### DIFF
--- a/src/node_sockaddr-inl.h
+++ b/src/node_sockaddr-inl.h
@@ -12,6 +12,8 @@
 
 namespace node {
 
+static constexpr uint32_t kLabelMask = 0xFFFFF;
+
 // Fun hash combine trick based on a variadic template that
 // I came across a while back but can't remember where. Will add an attribution
 // if I can find the source.
@@ -115,10 +117,9 @@ uint32_t SocketAddress::flow_label() const {
 }
 
 void SocketAddress::set_flow_label(uint32_t label) {
-  static constexpr uint32_t kMaxLabel = 1048575;
   if (family() != AF_INET6)
     return;
-  CHECK_LE(label, kMaxLabel);
+  CHECK_LE(label, kLabelMask);
   sockaddr_in6* in = reinterpret_cast<sockaddr_in6*>(&address_);
   in->sin6_flowinfo = label;
 }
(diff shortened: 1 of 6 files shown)


jasnell:
@addaleax ... thoughts on this one?

src/node_sockaddr-inl.h
```diff
@@ -107,6 +107,22 @@ int SocketAddress::port() const {
...
 
+uint32_t SocketAddress::flow_label() const {
+  if (family() != AF_INET6)
+    return 0;
+  const sockaddr_in6* in = reinterpret_cast<const sockaddr_in6*>(data());
+  return in->sin6_flowinfo;
+}
+
+void SocketAddress::set_flow_label(uint32_t label) {
+  static constexpr uint32_t kMaxLabel = 1048575;
```
addaleax: Maybe this could re-use the definition from `kMask` below?

src/quic/node_quic_crypto.cc
```diff
@@ -841,5 +841,50 @@ const char* crypto_level_name(ngtcp2_crypto_level level) {
...
+    const uint8_t* secret,
+    size_t secretlen) {
+  static constexpr uint32_t kMask = 0xFFFFF;
+  static constexpr size_t kInfoLen =
+      (sizeof(sockaddr_in6) * 2) + NGTCP2_MAX_CIDLEN;
+
+  uint8_t label[4] = {};
+
+  std::array<uint8_t, kInfoLen> plaintext;
+  size_t infolen = local.length() + remote.length() + cid.length();
```
addaleax:
```suggestion
  size_t infolen = local.length() + remote.length() + cid.length();
  CHECK_LE(infolen, kInfoLen);
```

src/quic/node_quic_crypto.cc
```diff
@@ -841,5 +841,50 @@ const char* crypto_level_name(ngtcp2_crypto_level level) {
...
+  ngtcp2_crypto_hkdf_expand(
+      label,
+      sizeof(label),
+      &ctx.md,
+      secret,
+      secretlen,
+      plaintext.data(),
+      infolen);
+
+  uint32_t ret = label[3] | label[2] << 8 | label[1] << 16 | label[0] << 24;
```
addaleax: Maybe use a `uint32_t` for `label`? Then we can skip the decoding step here

jasnell:
Landed

PR #303, opened 2020-01-24
quic: implement active connection id limit transport param

jasnell:
active_connection_id_limit was only recently added


doc/api/quic.md
```diff
@@ -1417,6 +1417,7 @@ added: REPLACEME
     `object.passphrase` if provided, or `options.passphrase` if it is not.
   * `activeConnectionIdLimit` {number}
   * `maxAckDelay` {number}
+  * `maxActiveCIDLimit` {number}
```
addaleax: Should we follow the ngtcp2 naming and spell `ConnectionID` out here? I think I’d be in favour of doing that, it’s a bit more intuitive

jasnell: Yeah I'm good with that

jasnell:
This was landed

PR #307, opened 2020-02-03
quic: fixup Debug statements to use new FPrintf

jasnell:


src/quic/node_quic_stream.cc
```diff
@@ -103,7 +103,7 @@ bool QuicStream::AddHeader(std::unique_ptr<QuicHeader> header) {
   }
 
   current_headers_length_ += header->length();
-  Debug(this, "Header - %s", header->ToString().c_str());
+  Debug(this, "Header - %s", header.get());
```
jasnell: btw, @addaleax, `header` here is a `std::unique_ptr<QuicHeader>` but the compiler complains that `ToString()` is not a member of `std::unique_ptr<QuicHeader>`. Very minor but worth looking at later.

addaleax: I guess you want to use `*header` here? `SPrintF` does not automatically dereference pointers

jasnell:
Landed

PR #309, opened 2020-02-03
quic: additional notes and cleanup

jasnell:


src/quic/node_quic_crypto.cc
```diff
@@ -81,6 +71,11 @@ bool DeriveTokenKey(
...
 }
 
+// Generates a random value that also contains a SHA256
+// hash of the random value. This is used within the
+// generation of a retry token. It would be equally
+// valid here just to fill buf with entirely random
+// bytes.
 void GenerateRandData(uint8_t* buf, size_t len) {
   uint8_t rand[16];
   uint8_t md[32];
```
addaleax: Uh … where is `md` filled in this function? Unless I’m missing something, it’s not filled at all (except maybe from a buffer overflow when filling `rand` with the digest), but it’s used for the output data?

I do feel like it makes sense to just use entirely random data instead of hashing random data.

jasnell: Yep, as I said in chat that was left over from when I adapted the original ngtcp2 example but it never really made much sense. Opening a PR that replaces it with an `EntropySource`

jasnell: https://github.com/nodejs/quic/pull/324

jasnell:
Landed

PR #320, opened 2020-02-04
quic: fixup segfault in StatsBase

jasnell:


src/quic/node_quic_util-inl.h
```diff
@@ -424,14 +424,6 @@ void StatsBase<T>::StatsMemoryInfo(MemoryTracker* tracker) const {
   tracker->TrackField("ack_histogram", ack_);
 }
 
-template <typename T>
-StatsBase<T>::~StatsBase() {
-  StatsBase<T>::StatsDebug stats_debug(static_cast<typename T::Base*>(this));
-  Debug(static_cast<typename T::Base*>(this),
-        "Destroyed. %s",
-        stats_debug.ToString().c_str());
```
addaleax: It might be nice to keep this as a function on `StatsBase` that is called from the individual destructors, rather than end up duplicating the code in each individual one?

src/quic/node_quic_session.h
```diff
@@ -625,8 +625,8 @@ class QuicApplication : public MemoryRetainer {
 // control. In other words, there's quite a bit going on within
 // a QuicSession object.
 class QuicSession : public AsyncWrap,
-                    public mem::NgLibMemoryManager<QuicSession, ngtcp2_mem>,
-                    public StatsBase<QuicSessionStatsTraits> {
+                    public StatsBase<QuicSessionStatsTraits>,
+                    public mem::NgLibMemoryManager<QuicSession, ngtcp2_mem> {
```
addaleax: Does this change anything? If it does, that’s probably a sign of (another?) bug

jasnell: oh, that was a leftover. I was trying it just in case it did.

src/quic/node_quic_util-inl.h
jasnell: Ah right, can give that a try

jasnell: Works! Updated

jasnell:
Landed

PR #321, opened 2020-02-04
[WIP] quic: fixup up client-connect-multiple-sequential

jasnell:
@addaleax or @danbev ... I can't remember exactly who added this test but it needed a bit of a fixup in order to successfully run but it's not clear if the fixup is testing the right thing or if there's another bug in here. Well, actually, I definitely think there's another bug in here :-) ... so definitely needs a lookover

### DIFF
--- a/test/parallel/test-quic-client-connect-multiple-sequential.js
+++ b/test/parallel/test-quic-client-connect-multiple-sequential.js
@@ -38,22 +38,28 @@ const { once } = require('events');
 
   const client = quic.createSocket({ client: { key, cert, ca, alpn: 'meow' } });
 
+  const reqs = [];
   for (const server of servers) {
     const req = client.connect({
       address: 'localhost',
       port: server.endpoints[0].address.port
     });
 
-    const [ stream ] = await once(req, 'stream');
-    stream.resume();
-    await once(stream, 'end');
-
-    server.close();
-    req.close();
-    await once(req, 'close');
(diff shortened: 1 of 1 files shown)


test/parallel/test-quic-client-connect-multiple-sequential.js
```diff
@@ -38,22 +38,28 @@ const { once } = require('events');
...
-    server.close();
-    req.close();
-    await once(req, 'close');
+    req.on('stream', common.mustCall((stream) => {
+      stream.resume();
+      stream.on('close', common.mustCall(() => {
+        req.close();
+      }));
+    }));
+    reqs.push(once(req, 'close'));
```
jasnell: If we change this line to `await once(req, 'close');` the test will hang.

addaleax:
The test was added by me in https://github.com/nodejs/quic/pull/176 … I intentionally provided it in two separate variants, one in which the requests are started in parallel and one where they are started sequentially.

This PR here would essentially un-do the point of having the sequential variant, so yes, this seems like it would hide a bug to me.

jasnell:
Yep, that's pretty much what I figured. There's something in the shutdown flow that's broken. I think it's also the source of the segfaults I've been seeing in various other tests. Will be chasing that down :-)

jasnell:
Closed in favor of https://github.com/nodejs/quic/pull/339

PR #326, opened 2020-02-05
quic: guard against use after destroy

jasnell:
Covers at least two of the known segfault cases...

Essentially: when a QuicSession is destroyed while the ngtcp2 callbacks are running, or while attempting to send packet data, we need to guard against use immediately after.

Adds additional UNLIKELY checks in the ngtcp2 callback handlers to make sure we abort the callbacks if the QuicSession is destroyed.


jasnell:
Landed

PR #327, opened 2020-02-05
quic: fixup test-quic-quicsocket

jasnell:
Avoid memory leak on ready listeners by moving all arg
validation before kMaybeBind call


jasnell:
Landed

PR #333, opened 2020-02-06
quic: fixup ipv6 only test + handle error case

jasnell:


test/parallel/test-quic-ipv6only.js
```diff
@@ -63,12 +60,13 @@ const kALPN = 'zzz';
   }));
 }
 
-// When the `ipv6Only` set to `true`, a client cann't connect to it
-// through "127.0.0.1".
+// // When the `ipv6Only` set to `true`, a client cann't connect to it
+// // through "127.0.0.1".
```
addaleax: …? ^^

jasnell: editing error. had commented out a section while debugging another ;-) fixed!

jasnell:
Landed

PR #336, opened 2020-02-08
src: clean up SocketAddress usage a bit

addaleax:
- Remove methods that duplicate functionality for slightly different
  signatures (e.g. taking a pointer vs taking a reference).
- Be more explicit about situations in which only IPv6 and IPv4 are
  supported.
- Use `operator==` for equality comparison.
- Remove unnecessary passing of `SocketAddress*` arguments.
- Othe minor cleanup.

##### Checklist


jasnell:
Landed

PR #337, opened 2020-02-10
quic: fix up early QuicSession destroy

jasnell:
Send a stateless connection close if destroying while still in
initial crypto level (there are no encryption keys at this point
so we can't send a regular connection close)


jasnell:
Landed

PR #338, opened 2020-02-11
quic: fix multiple segfaults and test-quic-maxconnectionsperhost

jasnell:


src/quic/node_quic_util-inl.h
```diff
@@ -42,11 +42,12 @@ size_t QuicCID::Hash::operator()(const QuicCID& token) const {
...
-    const QuicCID& lcid,
-    const QuicCID& rcid) const {
-  return lcid->datalen != rcid->datalen ?
-      false : memcmp(lcid->data, rcid->data, lcid->datalen) == 0;
+bool QuicCID::operator==(const QuicCID& other) const {
+  return memcmp(cid(), other.cid(), sizeof(ngtcp2_cid)) == 0;
+}
+
+bool QuicCID::operator!=(const QuicCID& other) const {
+  return !(*this == other);
```
jasnell: These weren't really related but since I was in here I went ahead and made the change

src/quic/node_quic_session.cc
```diff
@@ -2000,8 +2000,9 @@ void QuicSession::RemoveFromSocket() {
   }
 
   Debug(this, "Removed from the QuicSocket.");
-  socket_->RemoveSession(scid_, remote_address_);
+  QuicSocket* socket = socket_.get();
   socket_.reset();
+  socket->RemoveSession(scid_, remote_address_);
```
addaleax: I’m not sure I fully understand this bit … shouldn’t the behaviour be the same, except that with the early `.reset()` there is also the chance of a use-after-free on this line now?

src/quic/node_quic_util-inl.h
```diff
@@ -42,11 +42,12 @@ size_t QuicCID::Hash::operator()(const QuicCID& token) const {
   return hash;
 }
 
-bool QuicCID::Compare::operator()(
-    const QuicCID& lcid,
-    const QuicCID& rcid) const {
-  return lcid->datalen != rcid->datalen ?
-      false : memcmp(lcid->data, rcid->data, lcid->datalen) == 0;
+bool QuicCID::operator==(const QuicCID& other) const {
+  return memcmp(cid(), other.cid(), sizeof(ngtcp2_cid)) == 0;
```
addaleax: Does this mean that we assume a particular `datalen` now?

src/quic/node_quic_session.cc
jasnell: Use after free is exactly what was happening but I just thought of a better way to handle this :-) will update in a few mins

addaleax: Ah wait, use-after-free on the `QuicSession*` object? That’s… very subtle. :/ I think using a `BaseObjectPtr` for the `socket` variable here would be enough in that case?

src/quic/node_quic_util-inl.h
jasnell: the ngtcp2_cid struct uses a fixed size buffer now. The actual bits could be anything <= 20 bytes but the struct should be properly initialized.

src/quic/node_quic_session.cc
jasnell: Yeah, that's what I'll be changing it to :-) And yeah, it's super subtle. Took me a while to track this one down

jasnell: Ok updated to use a BaseObjectPtr now!

jasnell: But yeah, what was happening here is: the the QuicSocket maintains a map of QuicSession instances. The QuicSession maintains a pointer to the QuicSocket... when the QuicSession was being removed from the QuicSocket, it would free the QuicSession, but then try to use the socket_ reference from the QuicSession, and boom, crash when the object was garbage collected.

jasnell:
Landed

PR #341, opened 2020-02-12
quic: fix client migration and stream reset

jasnell:
Fixes two tests


jasnell:
@addaleax @nodejs/quic ... with this PR, all current tests are passing locally on Windows. It would be excellent to get some confirmation across platforms tho

jasnell:
Landed!! As of right this moment, all current QUIC tests are passing! Woo!

PR #342, opened 2020-02-13
quic: add tx packetloss test

jasnell:
This is functionally identical to the rx packetloss test, but with
random loss on transmitted packets


jasnell:
This was landed

PR #344, opened 2020-02-13
WTF?

jasnell:
After rebasing from nodejs/node, the build on Windows fails with a segfault when running node_mksnapshot... This commit fixes it but it's really not clear why. It's definitely something caused by the quic commit because without that commit the build works. Still trying to track it down. 

/cc @addaleax

### DIFF
--- a/src/node_options-inl.h
+++ b/src/node_options-inl.h
@@ -207,14 +207,20 @@ template <typename ChildOptions>
 void OptionsParser<Options>::Insert(
     const OptionsParser<ChildOptions>& child_options_parser,
     ChildOptions* (Options::* get_child)()) {
-  aliases_.insert(std::begin(child_options_parser.aliases_),
-                  std::end(child_options_parser.aliases_));
+  if (child_options_parser.aliases_.size() > 0) {
+    aliases_.insert(std::begin(child_options_parser.aliases_),
+                    std::end(child_options_parser.aliases_));
+  }
 
-  for (const auto& pair : child_options_parser.options_)
-    options_.emplace(pair.first, Convert(pair.second, get_child));
+  if (child_options_parser.options_.size() > 0) {
+    for (const auto& pair : child_options_parser.options_)
+      options_.emplace(pair.first, Convert(pair.second, get_child));
+  }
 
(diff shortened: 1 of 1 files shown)


jasnell:
Oh, the other thing weird is that the build does not fail when using `vcbuild experimental-quic debug`

addaleax:
So... unfortunately, I cannot reproduce this locally on Windows :/

I'm good with landing this, but maybe leave comments saying that these are workarounds that should eventually be removed...

jasnell:
So the build works just fine for you on Windows? Interesting! Which visual studio version are you running?

gengjiawen:
Works on windows too.
```console
PS D:\Developer\quic\Release> ./node -p process.versions
{
  node: '14.0.0-pre',      
  v8: '7.9.317.25-node.29',
  uv: '1.34.2',
  zlib: '1.2.11',
  brotli: '1.0.7',
  ares: '1.15.0',
  modules: '81',
  nghttp2: '1.40.0',       
  napi: '5',
  llhttp: '2.0.4',
  openssl: '1.1.1d',
  ngtcp2: '0.1.90',
  nghttp3: '0.1.90',
  cldr: '36.0',
  icu: '65.1',
  tz: '2019c',
  unicode: '12.1'
}
```

Env:
```console
  System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-4720HQ CPU @ 2.60GHz
    Memory: 6.18 GB / 15.89 GB
```

gengjiawen:
Also build passed on github windows CI. Maybe try a clean build.

jasnell:
Ok, very interesting... Upgrading my local environment to visual studio build tools 2019 "fixes" the build issue.

So, recap:

Sometime in the past few weeks there's been some change in nodejs/node master that conflicts with the changes in the QUIC PR. I'm thinking it most likely has to do with the order in which headers are init'd. This change is apparently triggering some as yet unknown msvc bug in visual studio build tools 2017. Bisecting the exact change that introduced the problem is going to be difficult because there are over 110 commits and the QUIC commits have to be cherry-picked at each bisect which is going to take forever. For now, the "fix" is to just build with VS 19.

PR #346, opened 2020-02-17
quic: use appropriate QuicEndpoint

jasnell:
Previously, send would always use the preferred endpoint. Use
the appropriate QuicEndpoint to send


jasnell:
Landed

PR #348, opened 2020-02-17
quic: multiple javascript cleanups and use validators

jasnell:
Multiple javascript cleanups including use of common validators


jasnell:
Landed

PR #349, opened 2020-02-19
quic: cleanup options and add highWaterMark/defaultEncoding

jasnell:
Allow passing in highWaterMark and defaultEncoding options
when creating new `QuicStream` instances using `openStream()`
and `pushStream()`, and allow setting the options for all
newly received peer-initiated `QuicStream` instances.

Additional cleanups and doc nits


jasnell:
Landed

PR #351, opened 2020-02-20
quic: configurable max early data

jasnell:
@addaleax @danbev @nodejs/quic

This is not ready to land as multiple tests fail. I'm still investigating those failures.

This PR does advance the ball forward a bit on 0RTT support but we're not quite there yet. There are a number of API design considerations that we need to address before we can completely support 0RTT.

First, an overview:

1. A 0RTT stream is a stream initiated by the client *as part of the initial handshake*. The key piece here is that the stream resource in ngtcp2 must be created before the handshake starts. Our API currently does not allow this. As soon as the client `QuicSession` is created, the handshake is started. User code does not have the opportunity to create the stream resource.

2. A 0.5RTT stream is a stream initiated by the client or server after the handshake has started but before it is finished. We currently do have the ability to create these but inside core.js currently, we defer creation of the stream resource until after the handshake completes. We will need to modify that behavior to allow earlier initiation.

3. A 1RTT stream is any stream created after the handshake completes. This is working just fine right now.

To support 0RTT streams, we either:

A. Defer the start of the TLS handshake until the user explicitly starts it. That is, rather than starting the handshake immediately upon creation of the client `QuicSession`, we make it so that the user must start it manually. Optionally under this approach we would have an `autoStart` option on creation of the client that would start the handshake immediately. This would give user code the opportunity to create the initial 0RTT stream. The api would be something like:

```js
// sessionTicket and remoteTransportParams are required for 0RTT early data
// and 0.5RTT data from the client. They resume the previous session.

const req = client.connect({ sessionTicket, remoteTransportParams, address, port })
const stream = req.openStream();
req.startHandshake();
```

If session resume is not being used (sessionTicket and remoteTransportParams are not provided) then the handshake would auto-start unless the `autoStart` option is explicitly `false`. 

A second option would be to initiate the 0RTT stream as part of the `connect()` operation. This is far more complicated with HTTP/3 because the headers need to be included also. The initiated 0RTT stream would be given to users in an event.

```js
const req = client.connect({
  sessionTicket,
  remoteTransportParams,
  stream: {
    headers: {
      ':method': 'get',
      ':path': '/',
      /* ... */
    }
  }
});

req.on('stream', (stream) => {
  if (stream.clientInitiated) {
    // this is the 0RTT stream
  } else {
    // these are push streams
  }
});
```

There may be other options here so if you have any ideas, please let me know.

### DIFF
--- a/doc/api/quic.md
+++ b/doc/api/quic.md
@@ -842,18 +842,6 @@ added: REPLACEME
 
 A `BigInt` representing the length of time the `QuicSession` was active.
 
-#### quicsession.earlyData
-<!-- YAML
-added: REPLACEME
--->
-
-* Type: {boolean}
-
-On server `QuicSession` instances, set to `true` on completion of the TLS
-handshake if early data is enabled. On client `QuicSession` instances,
-set to true on handshake completion if early data is enabled *and* was
-accepted by the server.
-
 #### quicsession.getCertificate()
 <!-- YAML
 added: REPLACEME
@@ -1130,6 +1118,18 @@ Initiates QuicSession key update.
 
 An error will be thrown if called before `quicsession.handshakeConfirmed`
 is equal to `true`.
+
+#### quicsession.usingEarlyData
+<!-- YAML
+added: REPLACEME
+-->
+
+* Type: {boolean}
+
+On server `QuicSession` instances, set to `true` on completion of the TLS
+handshake if early data is enabled. On client `QuicSession` instances,
+set to true on handshake completion if early data is enabled *and* was
+accepted by the server.
 
 ### Class: QuicClientSession extends QuicSession
 <!-- YAML
--- a/doc/api/quic.md
+++ b/doc/api/quic.md
@@ -842,6 +842,18 @@ added: REPLACEME
 
 A `BigInt` representing the length of time the `QuicSession` was active.
 
+#### quicsession.earlyData
+<!-- YAML
+added: REPLACEME
+-->
+
+* Type: {boolean}
+
+On server `QuicSession` instances, set to `true` on completion of the TLS
+handshake if early data is enabled. On client `QuicSession` instances,
+set to true on handshake completion if early data is enabled *and* was
+accepted by the server.
+
 #### quicsession.getCertificate()
 <!-- YAML
 added: REPLACEME
@@ -1153,9 +1165,8 @@ added: REPLACEME
 
 The `'sessionTicket'` event is emitted when a new TLS session ticket has been
 generated for the current `QuicClientSession`. The callback is invoked with
-three arguments:
+two arguments:
 
-* `sessionID` {Buffer} The serialized session ticket identifier.
 * `sessionTicket` {Buffer} The serialized session ticket.
 * `remoteTransportParams` {Buffer} The serialized remote transport parameters
   provided by the QUIC server.
@@ -1656,6 +1667,8 @@ added: REPLACEME
     error will be thrown. It is strongly recommended to use 2048 bits or larger
     for stronger security. If omitted or invalid, the parameters are silently
     discarded and DHE ciphers will not be available.
+  * `earlyData` {boolean} Set to `false` to disable 0RTT early data.
+    Default: `true`.
   * `ecdhCurve` {string} A string describing a named curve or a colon separated
     list of curve NIDs or names, for example `P-521:P-384:P-256`, to use for
     ECDH key agreement. Set to `auto` to select the
--- a/src/quic/node_quic_crypto.cc
+++ b/src/quic/node_quic_crypto.cc
@@ -35,6 +35,21 @@ using v8::Value;
 
 namespace quic {
 
+bool SessionTicketAppData::Set(const uint8_t* data, size_t len) const {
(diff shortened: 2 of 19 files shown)


jasnell:
Ok, so this turned into a much larger change. In order to properly enable early data, I had to revamp a bunch of the state tracking across the various objects and took the opportunity to also clean things up. Early data is still not full tested but pending QuicStream is much more reliable and clearer now than it was previously. I will be adding a 0RTT test soon

jasnell:
Still working through this trying to get 0RTT and 0.5RTT early data to work. The timing is pretty tricky if we want it to have a reasonable developer experience. Right now, there are two issues that I likely won't get to until either this weekend or early next week:

1. Moving the server side `onSessionReady` to occur once the client initial handshake has started makes it too late to add listeners that impact the handshake, so adding keylog and OCSP request event handlers aren't called. Need to rework the timing on these a bit.

2. On the client side, when `autoStart` is `false`, the handshake is kicked off by calling `write()` or `end()` on the `QuicStream`. It shouldn't be.

jasnell:
@addaleax this is ready for review but it does not get 0RTT early data fully functional. All of the mechanism is in place but there are still some timing issues getting in the way. It does make a number of improvements that need to land regardless.

jasnell:
Getting this closer to supporting 0RTT but just can't seem to get ngtcp2 to actually do it. Giving up for the day and going to try a few other things tomorrow. In the meantime, there's a bunch in this PR that should land and a few more improvements coming.

doc/api/quic.md
```diff
@@ -842,6 +842,18 @@ added: REPLACEME
 
 A `BigInt` representing the length of time the `QuicSession` was active.
 
+#### quicsession.earlyData
```
addaleax: Can we name this (and the option) something like `hasEarlyData`, `allowsEarlyData`, or similar, so that it sounds more boolean-y?

src/quic/node_quic_crypto.cc
```diff
@@ -739,19 +802,39 @@ void InitializeTLS(const QuicSession& session) {
...
     crypto::SecureContext* sc,
+    bool early_data,
     ngtcp2_crypto_side side) {
-  constexpr auto ssl_server_opts =
-      (SSL_OP_ALL & ~SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS) |
-      SSL_OP_SINGLE_ECDH_USE |
-      SSL_OP_CIPHER_SERVER_PREFERENCE |
-      SSL_OP_NO_ANTI_REPLAY;
+  // TODO(@jasnell): Should this be a static value?
+  constexpr static unsigned char session_id_ctx[] = "node.js quic server";
```
addaleax: My reading of the `SSL_CTX_set_session_id_context` is that this should be fine, but feel free to wait for confirmation from more people more familiar with crypto/tls :)

src/quic/node_quic_crypto.h
```diff
@@ -130,6 +133,37 @@ v8::Local<v8::Value> GetALPNProtocol(const QuicSession& session);
...
+    STATUS_NONE,
+    STATUS_RENEW
+  };
+
+  explicit SessionTicketAppData(SSL_SESSION* session) : session_(session) {}
+  bool Set(const uint8_t* data, size_t len) const;
+  bool Get(uint8_t** data, size_t* len) const;
+
+ private:
+  mutable bool set_ = false;
```
addaleax: This should not be mutable (and `Set()` should not be const), I think? Calling `Set()` does change the internal state of this object in a visible way.

src/quic/node_quic_session.cc
```diff
@@ -66,11 +66,14 @@ void SetConfig(Environment* env, int idx, uint64_t* val) {
...
   QuicSession* session = static_cast<QuicSession*>(user_data);
+  if (!UNLIKELY(session->env()->debug_enabled(DebugCategory::NGTCP2_DEBUG)))
+    return;
   va_list ap;
   va_start(ap, fmt);
   std::string format(fmt, strlen(fmt) + 1);
   format[strlen(fmt)] = '\n';
-  Debug(session->env(), DebugCategory::NGTCP2_DEBUG, format, ap);
+  // TODO(@jasnell): Debug() currently is not working with va_list here.
+  vfprintf(stderr, format.c_str(), ap);
```
addaleax: I think using vprintf is just fine – this callback is only every called if debugging is enabled anyway, right?

src/quic/node_quic_crypto.cc
jasnell: For now, I'll leave it here but with a better todo message

jasnell:
> Just an idea, but maybe we could delay sending the first data with a SetImmediate() so that any data written in the same tick as the .connect() call could use 0RTT? Then again I’m not sure if the extra complexity is worth that…

I'd much rather avoid introducing the set immediate. I believe the current state of this is very close to what we need, I just have to trouble shoot exactly why it's not sending off the early data when it should (as far as I know the criteria for doing so are being met but ngtcp2 is still choosing not too).

jasnell:
Landed. Will continue the work on 0RTT in a new PR

PR #352, opened 2020-02-24
quic: fix clang build

oyyd:
It seems that clang doesn't implicitly convert "unsigned long"
into "unsigned long long" for function templates.

##### Checklist


jasnell:
Landed

PR #366, opened 2020-04-11
quic: cleanups and use backing store for statsbase

jasnell:
Two commits here:

The first makes a number of style cleanups.

The second updates StatsBase to use a BackingStore instead of the aliased buffer. If https://github.com/nodejs/node/pull/32778 lands, it will be updated to use the `AliasedStruct` added there.


src/quic/node_quic_util-inl.h
```diff
@@ -42,6 +42,12 @@ size_t QuicCID::Hash::operator()(const QuicCID& token) const {
   return hash;
 }
 
+QuicCID& QuicCID::operator=(const QuicCID& cid) {
+  if (this == &cid) return *this;
+  this->~QuicCID();
+  return *new(this) QuicCID(std::move(cid));
```
addaleax, line 48: Something is off about this – the `std::move` here won’t do anything to a const reference, so either you meant to implement a move operator rather than a copy operator, or the `std::move` can be dropped?

src/quic/node_quic_util-inl.h
```diff
@@ -323,23 +330,29 @@ template <typename T>
...
-  // TODO(@jasnell): The follow are checks instead of handling
-  // the error. Before this code moves out of experimental,
-  // these should be change to properly handle the error.
 
-  wrap->DefineOwnProperty(
+  // Create the backing store for the statistics
+  size_t size = sizeof(typename T::Stats);
+  size_t count = size / sizeof(uint64_t);
+  stats_store_ = v8::ArrayBuffer::NewBackingStore(env->isolate(), size);
+  stats_ = new (stats_store_->Data()) typename T::Stats;
```
addaleax, line 340: Not sure if it makes a difference right now, but generally there should be a call to `~typename T::Stats()` when placement new is being used (probably from ~StatsBase).

You should also be able to do something like `typedef typename T::Stats Stats;` in the class definition, so that you don’t always have to spell out `typename T::Stats`.

src/quic/node_quic_util-inl.h
jasnell: Actually, I don't think we're actually using this at all so likely best to just remove it.

danbev:
Sorry, I should have landed this before 0eed19b83ccb3b170934d4729762d1c0c6a5ed43 :disappointed: 
@jasnell Would you be able to rebase?

jasnell:
No worries at all @danbev ... will rebase before landing.

jasnell:
Landed

PR #369, opened 2020-04-15
quic: use OneByteString from util in quic_session

danbev:
This commit replaces two usages of `v8::String::NewFromOneByte` with the
`OneByteString` function from util.h.

##### Checklist


danbev:
Landed in 80ac6aa9542a53ee756f465d36ac7f660b1e56b3.

PR #372, opened 2020-04-16
quic: remove namespace qualifier for SecureContext

danbev:
Commit 0eed19b83c ("quic: remove namespace qualifier for SecureContext")
removed these in `node_quic_session.cc`, but I missed that there were the
same issues in these two files as well.

##### Checklist


jasnell:
Landed

PR #373, opened 2020-04-20
test: use gtest ASSERT_EQ instead of CHECK_EQ

danbev:
This commit changes the `CHECK_EQ` in test_quic_buffer.cc to use Google
Test's `ASSERT_EQ` instead. An ASSERT will stop processing the test like
CHECK_EQ will so the behaviour should be the same.

This can be run locally using (there is currently a failure in the
test_quic_cid.cc):
```console
$ ./out/Release/cctest --gtest_filter=QuicBuffer.*
```

##### Checklist


jasnell:
Landed

PR #377, opened 2020-04-24
quic: revert validators.js changes

jasnell:
There was pushback in nodejs/node on altering the validators
to use the allowUndefined option so revert that here. Not
going to fight that battle.


jasnell:
Oy, totally forgot about the test updates. That's what I get for being in a rush. Will add those and the lint fixes later on today

danbev:
Landed in 29faf3d9b132aac28e0187ef3a4190bc51dbe030.

PR #380, opened 2020-04-30
quic: use using declaration instead of explicit ns

danbev:
`node_quic_http3_application.cc` has a few v8 specific using declarations
but they are currently not used and instead explicit namespacing is
used. This commit updates the code to use the using declarations and
also removes the ones that are no longer in use.

##### Checklist


danbev:
Landed in d2ed3a8f308995ae5687ab80f2e18160e0a4c488.

Issue #386, opened 2020-05-15
Memory Fragmentation (WIP)


splitice:
* **Version**:  latest / 5baab3f3a05548d3b51bea98868412b08766e34d
* **Platform**: Linux Debian ARMv7

### What steps will reproduce the bug?

Long running QUIC client making regular bidirectional streams in both directions

### How often does it reproduce? Is there a required condition?

24-48hrs for maximum effect

### What is the expected behavior?

Memory sitting around 20-30MB (max old size of 30MB). This is the usage of the process without the QUIC server components compiled in and using TCP sockets for the same protocol in place of QUIC.

### What do you see instead?

Memory peaking at 100MB+ (OOM on test device)

### Additional information

QUIC (most likely ngtcp2 allocated memory) seems to create a high rate of memory fragmentation when using the default glibc malloc in real world conditions. While 100MB of ram is likely not an issue in server applications in the low-end space this is significant. Additionally as this ram is allocated with an external allocator / not v8 it is not allocated as part of the nodejs memory pool it will grow unrestricted by parameters such as max old space size.

The usage appears to be fragmentation rather than a leak, however I have not entirely ruled that out.

Currently I'm testing using jemalloc to see if it exhibits more sane behavior (results to following in coming week). Turn around time on replication means that I'll be testing this over the coming week. If it jemalloc results in sane memory usage then that supports a fragmentation situation. So far with a runtime of 2hrs this seems to be supported.

jasnell:
Loving these issues. Keep them coming if you can. Will be digging back in on the quic code early next week.

splitice:
jemalloc with narenas=1 shows none of the excessive growth (confirmed over 48hrs).

That supports this being fragmentation over a leak.

I'm not really sure what the next steps here are.

jasnell:
We'll need to reproduce the issue. We're not doing much special around allocations here so it's going to take a bit to figure out and nail down. Just having as much information as possible on your test case and data you've collected would be a great start

splitice:
@jasnell do you have a generic server and client example that exercises the streams api (say opening a stream and reading and writing for example)? Perhaps leaving something like that running for a few days would be a good test.

This may be just a case of glibc being derpy and fragmenting with smaller allocations. Normally with TCP based streams this wouldn't be seen but I'm imagining there are alot of small allocations for dealing with small signalling packets in QUIC.

jasnell:
[...]

PR #388, opened 2020-05-18
quic: hopefully avoid memory fragmentation issue

jasnell:
@splitice ... I'm hoping this provides a reasonable fix for  https://github.com/nodejs/quic/issues/386

Previously, `QuicPacket` was allocating an `std::vector<uint8_t>` of `NGTCP2_MAX_PKT_SIZE` bytes, then the packet would be serialized into the buffer, and the `std::vector` would be resized based on the number of bytes serialized. I suspect the memory fragmentation that you're seeing is because of those resize operations not freeing memory in chunks that are aligned with the allocation. This changes `QuicPacket` to use a stack allocation that is always `NGTCP2_MAX_PKT_SIZE` bytes and the size of the serialized packet is just recorded without any resizing. When the memory is freed now, it should be freed in large enough chunks to cover subsequent allocations.

This is the most likely culprit in the current implementation so let me know if this patch fixes it. Will keep this open until I get confirmation from you one way or the other.

##### Checklist

### DIFF
--- a/src/quic/node_quic_socket.h
+++ b/src/quic/node_quic_socket.h
@@ -151,11 +151,11 @@ class QuicPacket : public MemoryRetainer {
 
   QuicPacket(const char* diagnostic_label, size_t len);
   QuicPacket(const QuicPacket& other);
-  uint8_t* data() { return data_.data(); }
-  size_t length() const { return data_.size(); }
+  uint8_t* data() { return data_; }
+  size_t length() const { return len_; }
   uv_buf_t buf() const {
     return uv_buf_init(
-      const_cast<char*>(reinterpret_cast<const char*>(data_.data())),
+      const_cast<char*>(reinterpret_cast<const char*>(&data_)),
       length());
   }
   inline void set_length(size_t len);
@@ -166,7 +166,8 @@ class QuicPacket : public MemoryRetainer {
   SET_SELF_SIZE(QuicPacket);
 
  private:
-  std::vector<uint8_t> data_;
+  uint8_t data_[NGTCP2_MAX_PKTLEN_IPV4];
+  size_t len_ = NGTCP2_MAX_PKTLEN_IPV4;
   const char* diagnostic_label_ = nullptr;
 };
 
--- a/src/quic/node_quic_socket-inl.h
+++ b/src/quic/node_quic_socket-inl.h
@@ -18,6 +18,7 @@ namespace quic {
 std::unique_ptr<QuicPacket> QuicPacket::Create(
     const char* diagnostic_label,
     size_t len) {
+  CHECK_LE(len, NGTCP2_MAX_PKTLEN_IPV4);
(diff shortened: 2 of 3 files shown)


splitice:
Thanks @jasnell. I've cherry-picked built and installed it. I'll aim to leave the hardware running with glibc malloc and your commit over the weekend for report on Monday.

addaleax:
> I suspect the memory fragmentation that you're seeing is because of those resize operations not freeing memory in chunks that are aligned with the allocation.

@jasnell You’re 100 % correct here – `std::vector::resize()` *never* frees memory. If that is the expectation here, you should call `std::vector::shrink_to_fit()` after it. I suspect that that would be the right fix here if that’s what’s causing the problems… this PR in its current state should only ever increase memory usage because it makes `QuickPacket` never use less than the maximum amount of data?

jasnell:
> this PR in its current state should only ever increase memory usage because it makes QuickPacket never use less than the maximum amount of data?

The `QuicPacket` instances are transient and exist only as long as it takes to serialize the complete the UDP send, then they are freed as quickly as possible after the udp send completes. I want to refactor this a bit more to allow us to size them more explicitly when we know we aren't going to need the full 1252 bytes. Because of coalescing requirements, we don't know in advance how much of the packet ngtcp2 is going to use every time we ask it to serialize.

splitice:
@jasnell This seems to work well. Currently sitting at 55MB after 3 days.

src/quic/node_quic_socket.h
```diff
@@ -166,7 +166,8 @@ class QuicPacket : public MemoryRetainer {
   SET_SELF_SIZE(QuicPacket);
 
  private:
-  std::vector<uint8_t> data_;
+  uint8_t data_[NGTCP2_MAX_PKTLEN_IPV4];
```
lundibundi, line 169: The PR message talks about using `NGTCP2_MAX_PKT_SIZE` but new code uses `NGTCP2_MAX_PKTLEN_IPV4` is that correct? (note that the check in constructor `QuicPacket::QuicPacket(const char* diagnostic_label, size_t len)` still checks for `NGTCP2_MAX_PKT_SIZE`)

Also, `NGTCP2_MAX_PKT_SIZE` recently changed to `NGTCP2_DEFAULT_MAX_UDP_PAYLOAD_SIZE` (https://github.com/ngtcp2/ngtcp2/commit/38feb91ee1e0556696434556923bcff7093ea6cd)

jasnell: I'll double check on that. May have mentally mixed those up

jasnell:
> This seems to work well. Currently sitting at 55MB after 3 days.

Fantastic. Will get this updated in the next day or two and get it landed.

jasnell:
Closing in favor of https://github.com/nodejs/node/pull/33912

